PR #58, opened 2026-02-24
feat: make wallet password optional (agent-friendly)

askeluv:
## Problem

AI agents create wallets interactively, pick a password, then lose it on session restart. This is a fundamental issue for agent-first CLIs — agents are amnesiac by nature.

## Solution

Make wallet passwords **optional by default**. Keys are stored unencrypted (like `~/.ssh/` keys). Users who want encryption at rest opt in via `NANSEN_WALLET_PASSWORD` env var.

### How it works

- `nansen wallet create my-wallet` → no password, keys stored as plaintext JSON
- `NANSEN_WALLET_PASSWORD=xyz nansen wallet create my-wallet` → encrypted with that password
- All other commands (export, delete, execute, send) auto-detect whether the wallet is password-protected
- If existing wallets have a password, new wallets require the env var too (consistency)
- No interactive password prompts anywhere in wallet create

### Security model

Same as SSH keys — the filesystem is the security boundary. Agents running in sandboxed containers (Docker, cloud VMs) get isolation from the runtime. Humans who want defense-in-depth set the env var in their shell profile or secrets manager.

## Changes

- `src/wallet.js`: Optional password in create, export, delete; plaintext encrypt/decrypt path
- `src/trading.js`: Auto-detect passwordless wallets in execute flow
- `src/transfer.js`: Skip password verification for unprotected wallets
- `src/x402.js`: Handle passwordless wallet flow

## Tests

All 135 existing tests pass (wallet: 25, trading: 55, transfer: 55).

### DIFF
--- a/src/__tests__/wallet.test.js
+++ b/src/__tests__/wallet.test.js
@@ -107,7 +107,7 @@ describe('encryption', () => {
     const password = 'testpassword123';
     const encrypted = encryptKey(key, password);
     
-    expect(() => decryptKey(encrypted, null)).toThrow('Incorrect password');
+    expect(() => decryptKey(encrypted, null)).toThrow('Wallet is encrypted. Set NANSEN_WALLET_PASSWORD.');
   });
 
   it('decryptKey should reject plaintext with encryption metadata', () => {
@@ -307,6 +307,20 @@ describe('wallet CRUD', () => {
     const solPub = Buffer.from(exported.solana.privateKey.slice(64), 'hex');
     
     expect(base58Encode(solPub)).toBe(result.solana);
+  });
+
+  it('should round-trip passwordless wallet: create, export, delete', () => {
+    const result = createWallet('nopass', null);
+    expect(result.evm).toBeDefined();
+    expect(result.solana).toBeDefined();
+
+    const exported = exportWallet('nopass', null);
+    expect(exported.evm.privateKey).toBeDefined();
+    expect(exported.solana.privateKey).toBeDefined();
+
+    deleteWallet('nopass', null);
+    const list = listWallets();
+    expect(list.wallets.find(w => w.name === 'nopass')).toBeUndefined();
   });
 
   it('should update default after deleting default wallet', () => {
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -131,7 +131,7 @@ export function decryptKey(encryptedData, password) {
 
   // For encrypted data, password is required
   if (!password) {
-    throw new Error('Incorrect password');
+    throw new Error('Wallet is encrypted. Set NANSEN_WALLET_PASSWORD.');
   }
 
   const salt = Buffer.from(encryptedData.salt, 'hex');
@@ -289,6 +289,9 @@ async function promptPassword(question, deps = {}) {
   const promptFn = deps.promptFn;
   if (promptFn) {
     return promptFn(question, true);
+  }
+  if (!process.stdin.isTTY) {
+    throw new Error('Wallet is encrypted. Set NANSEN_WALLET_PASSWORD.');
   }
   // Fallback to readline
   const rl = readline.createInterface({ input: process.stdin, output: process.stdout });
@@ -540,9 +543,6 @@ export function buildWalletCommands(deps = {}) {
             exit(1);
             return;
           }
-
-
-
           if (!password) {
             console.error('⚠️  No NANSEN_WALLET_PASSWORD set — private keys will be stored unencrypted.');
           }
@@ -558,6 +558,7 @@ export function buildWalletCommands(deps = {}) {
             log(`    Base (recommended, lower fees): send USDC to ${result.evm}`);
             log(`    Solana: send USDC to ${result.solana}`);
             log('');
+            log('  ⚠️  This is a hot wallet. Do not deposit more than you can afford to lose.');
             return;
           } catch (err) {
             log(`❌ ${err.message}`);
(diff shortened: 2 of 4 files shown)


src/wallet.js
```diff
@@ -78,6 +78,11 @@ function deriveKey(password, salt) {
  * Returns a JSON-serializable object with all params needed for decryption.
  */
 export function encryptKey(privateKeyHex, password) {
+  // No password → store key as plaintext wrapper
+  if (!password) {
```
0xlaveen, line 82: Nit: the field name `plaintext` is a bit on-the-nose — if someone `cat`s the wallet file, it literally advertises "here is the plaintext key." SSH doesn't label key files as "PLAINTEXT PRIVATE KEY" either.

Consider a less descriptive name like `key` or `data`, with a sibling marker:
```js
if (!password) {
  return { data: privateKeyHex, encrypted: false };
}
```

Not a blocker, just security hygiene.

src/x402.js
```diff
@@ -96,18 +96,21 @@ export async function* createPaymentSignatures(response, url, options = {}) {
...
-
-  let exportWallet, listWallets;
+  let password = options.password || process.env.NANSEN_WALLET_PASSWORD || null;
+  // For passwordless wallets, we still need to proceed
+  let walletMod;
   try {
-    const walletMod = await import('./wallet.js');
-    exportWallet = walletMod.exportWallet;
-    listWallets = walletMod.listWallets;
+    walletMod = await import('./wallet.js');
```
0xlaveen: The restructure here is slightly fragile — `exportWallet` and `listWallets` are destructured after the try/catch, but if the import fails it already returns early. Can simplify to a single block:

```js
try {
  const { exportWallet, listWallets, getWalletConfig } = await import('./wallet.js');
  const config = getWalletConfig();
  if (config.passwordHash && !password) return;
  // ... use exportWallet, listWallets directly below
} catch { return; }
```

Avoids the dangling `let walletMod` and the separate re-assignment.

Codier:
@claude Please review this PR and provide your assessment.

src/wallet.js
```diff
@@ -78,6 +78,11 @@ function deriveKey(password, salt) {
  * Returns a JSON-serializable object with all params needed for decryption.
  */
 export function encryptKey(privateKeyHex, password) {
+  // No password → store key as unencrypted wrapper
+  if (!password) {
```
Codier, line 82: 🔴 **Blocking — Security**: `decryptKey` trusts `encryptedData.encrypted === false` without any other validation. If an attacker can modify the wallet JSON file on disk (or if a bug writes a malformed file), they could set `encrypted: false` on an *actually encrypted* wallet and `data` to an arbitrary key, bypassing encryption entirely.

Consider adding a secondary check — e.g. store a HMAC of the plaintext key alongside it so you can at least verify integrity, or validate that `encryptedData` has the expected shape for each path (plaintext must not have `salt`/`iv`/`authTag`, and vice versa).

src/wallet.js
```diff
@@ -400,7 +416,7 @@ export function exportWallet(name, password) {
   }
 
   const config = getWalletConfig();
-  if (!verifyPassword(password, config)) {
```
Codier, line 403: 🟡 **Should Fix**: The `⚠️ No NANSEN_WALLET_PASSWORD set` warning goes to `log()` which is stdout. For structured JSON output (hard rule), this warning string will break JSON parsers. The wallet creation result JSON should include a `"warning"` field instead, and the emoji warning should go to stderr.

src/wallet.js
```diff
@@ -103,6 +108,11 @@ export function encryptKey(privateKeyHex, password) {
  * @throws {Error} If password is wrong
  */
```
Codier, line 109: 🟡 **Should Fix**: The CLI handler checks `config.passwordHash && !password` and exits, but the `createWallet()` function itself does NOT enforce this. Any direct caller of `createWallet(name, null)` will happily create an unencrypted wallet even when other wallets are encrypted. The guard should live in the function, not just the CLI handler.

src/__tests__/wallet.test.js
```diff
@@ -92,6 +92,16 @@ describe('encryption', () => {
 
     expect(() => decryptKey(encrypted, 'wrong-password')).toThrow('Incorrect password');
   });
+
+  it('encryptKey with null password returns unencrypted wrapper', () => {
+    const result = encryptKey('deadbeef', null);
+    expect(result).toEqual({ data: 'deadbeef', encrypted: false });
+  });
+
+  it('decryptKey handles unencrypted wrapper', () => {
```
Codier, line 100: 🟢 **Nitpick**: Good that you added tests for the null-password path. Consider also testing the negative case: `decryptKey` called with an encrypted blob but null password should still throw, and calling with `{encrypted: false, data: ...}` that also has `salt`/`iv` fields should probably be rejected.

src/x402.js
```diff
@@ -96,14 +96,14 @@ export async function* createPaymentSignatures(response, url, options = {}) {
   const ranked = rankRequirements(requirements);
   if (ranked.length === 0) return;
 
-  const password = options.password || process.env.NANSEN_WALLET_PASSWORD;
-  if (!password) return;
-
+  const password = options.password || process.env.NANSEN_WALLET_PASSWORD || null;
   let exportWallet, listWallets;
```
Codier: 🟡 **Should Fix**: When `config.passwordHash && !password`, you silently `return` — the payment just fails without any signal to the caller. At minimum yield a diagnostic or log to stderr so the user knows *why* x402 payment was skipped.

Codier:
Overall having no password for the wallet feels like a security risk here. I would really hope to get @mrh-nansen to at least approve the overall concept before moving further on this PR

PR #73, opened 2026-02-25
fix: remove credit-spending validation from login, add non-TTY support

TimNooren:
## Problem
`nansen login` validates the API key by calling `tokenScreener`, which:
1. **Spends a credit** on metered plans
2. Can fail with `Payment required` due to x402 gating

Additionally, agents running in non-TTY environments hang indefinitely waiting for interactive input.

## Changes
- **Remove validation call** — save the API key directly without calling `tokenScreener`. Bad keys are caught naturally on the first real command (401 error with clear message).
- **Non-TTY detection** — when `!process.stdin.isTTY` and no `--api-key` flag provided, check `NANSEN_API_KEY` env var as fallback. If neither available, exit with a clear error instead of hanging.

## Agent-friendly behavior
```bash
# Works (flag)
nansen login --api-key sk-xxx

# Works (env var in non-TTY)
NANSEN_API_KEY=sk-xxx nansen login

# Clear error (non-TTY, no key)
nansen login
# ❌ No API key provided. Use: nansen login --api-key <key>
#    Or set NANSEN_API_KEY environment variable.
```

## Tests
All 575 tests passing. Updated login tests to reflect no-validation behavior and added non-TTY test case.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -1010,7 +1010,8 @@ export function buildCommands(deps = {}) {
     saveConfigFn = saveConfig,
     deleteConfigFn = deleteConfig,
     getConfigFileFn = getConfigFile,
-    exit = process.exit
+    exit = process.exit,
+    isTTY = process.stdin.isTTY
   } = deps;
 
   return {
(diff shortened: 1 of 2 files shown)


src/cli.js
```diff
@@ -1019,39 +1020,38 @@ export function buildCommands(deps = {}) {
...
         log('\n❌ No API key provided');
         exit(1);
         return;
       }
-      
-      // Validate the key with a test request
-      log('\nValidating API key...');
-      try {
-        const testApi = new NansenAPIClass(apiKey.trim());
-        await testApi.tokenScreener({ chains: ['solana'], pagination: { page: 1, per_page: 1 } });
```
kome12, line 1039: Maybe instead of calling a credit consuming endpoint we can make a new endpoint to "test" api keys for 0 credits? Then users will at least know during the set up if their API key is incorrect

TimNooren: Maybe we can consider this a nice to have..? (Or at least kind of independent from this PR so I can merge it😅). The more important thing I think is to not return a 402 if the user passes an invalid key so the user is informed on the first actual call they try to make (the cli enforces the correct `apikey` header,  so we can easily catch this in the API).

kome12: Makes sense, approved!

PR #82, opened 2026-02-26
fix: prevent scientific notation for large token amounts

### LINKED ISSUE #78: Bug: Scientific notation breaks trading amounts ≥1000 tokens
## Description

When token amounts hit ≥10²¹ raw units (e.g. 1000 tokens with 18 decimals), `JSON.parse()` converts them to JS numbers, then `String()` produces `"1e+21"` which the trading API rejects.

## Steps to Reproduce

1. Try to trade 1000 of any 18-decimal token (e.g. DEGEN on Base)
2. The raw amount is `1000000000000000000000` (10²¹)
3. CLI sends `"1e+21"` as the amount string → API rejects it

## Expected Behavior

The CLI should handle arbitrarily large token amounts without scientific notation. Use `BigInt` or string-based arithmetic for raw amounts.

## Impact

Any user trading ≥1000 of a standard 18-decimal token hits this. Very common scenario.

## Current Workaround

Keep trade amounts under 1000 tokens (≤999) to stay below 10²¹ raw units.

Codier:
## Fix

Closes #78

**Problem:** When token amounts >= 10²¹ raw units (e.g. 1000 tokens with 18 decimals), `JSON.parse()` in `parseArgs()` converts the CLI string to a JS Number. `String()` then produces `"1e+21"` which the API rejects.

**Solution:** Keep numeric CLI argument strings as-is instead of parsing them through `JSON.parse()`. Only use parsed values for non-number types (objects, arrays, booleans). Explicitly `parseInt()` the few options that need actual numbers (`retries`, `cache-ttl`).

**Changes:**
- `src/cli.js` — `parseArgs()`: skip `JSON.parse` for numeric strings
- `src/cli.js` — `runCLI()`: explicit parseInt for retries/cache-ttl
- Tests updated + 2 new test cases for large amounts

**All 605 tests pass.**

### Repro
```bash
nansen quote --chain base --from 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee --to 0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913 --amount 1000000000000000000000
```
Before: `Error: Amount must not use scientific notation`
After: Amount passed correctly as string

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -1604,12 +1604,13 @@ export async function runCLI(rawArgs, deps = {}) {
     // Configure retry options
     const retryOptions = flags['no-retry'] 
       ? { maxRetries: 0 } 
-      : { maxRetries: options.retries !== undefined ? parseInt(options.retries, 10) : 3 };
+      : { maxRetries: options.retries !== undefined ? (Number.isNaN(parseInt(options.retries, 10)) ? 3 : parseInt(options.retries, 10)) : 3 };
     
     // Configure cache options
+    const cacheTtl = options['cache-ttl'] !== undefined ? parseInt(options['cache-ttl'], 10) : 300;
     const cacheOptions = {
       enabled: flags['cache'] && !flags['no-cache'],
-      ttl: options['cache-ttl'] !== undefined ? parseInt(options['cache-ttl'], 10) : 300
+      ttl: Number.isNaN(cacheTtl) ? 300 : cacheTtl
     };
     
     const defaultHeaders = {};


src/cli.js
```diff
@@ -245,9 +245,12 @@ export function parseArgs(args) {
       if (key === 'pretty' || key === 'help' || key === 'version' || key === 'table' || key === 'no-retry' || key === 'cache' || key === 'no-cache' || key === 'stream' || key === 'enrich') {
         result.flags[key] = true;
       } else if (next && !next.startsWith('-')) {
-        // Try to parse as JSON first
+        // Try to parse as JSON first (for objects/arrays/booleans),
+        // but keep numeric strings as strings to avoid precision loss
+        // and scientific notation for large integers (e.g. 1e+21).
```
Codier, line 250: 🟢 **Nitpick**: `parseInt` silently returns `NaN` for garbage input (e.g. `--retries foo`). Consider adding a guard or fallback: `const r = parseInt(options.retries, 10); retryOptions.maxRetries = Number.isNaN(r) ? 3 : r;`

Codier: Good catch — added NaN guards with fallback to defaults (3 for retries, 300 for cache-ttl). Pushed in 3d90d63.

PR #85, opened 2026-02-26
fix: auto-substitute wrapped native token with native sentinel

### LINKED ISSUE #80: Bug: WETH address as --from causes TRANSFER_FROM_FAILED when user has native ETH
## Description

On Base, using the WETH address (`0x4200000000000000000000000000000000000006`) as `--from` causes `TRANSFER_FROM_FAILED` because the wallet holds native ETH, not WETH. Users need to know to use the native ETH address (`0xEeEeEeEeEeEeEeEeEeEeEeEeEeEeEeEeEeEeEeEe`) instead.

## Steps to Reproduce

1. Have native ETH in wallet on Base
2. Run a swap with `--from 0x4200000000000000000000000000000000000006` (WETH)
3. Transaction fails with `TRANSFER_FROM_FAILED`

## Expected Behavior

Either:
- Auto-detect that user has native ETH but not WETH, and use the native address automatically
- Show a clear error: "You have 0.05 ETH but 0 WETH — did you mean native ETH?"
- Document the distinction clearly in help text

## Impact

Confusing for anyone who looks up ETH on Base and finds the WETH contract address.

Codier:
## Summary
- When `--from` is a wrapped native token (WETH on Ethereum/Base, WBNB on BSC), checks on-chain balances via RPC before requesting a quote
- If wallet has sufficient wrapped token balance, proceeds normally with no changes
- If wallet lacks the wrapped token but has sufficient native ETH/BNB, auto-substitutes `--from` with the native token sentinel (`0xeee...eee`) and informs the user
- If wallet has neither, proceeds as-is and lets the API handle the error
- Only applies to `--from` (input token); `--to` is left unchanged

Closes #80

## Test plan
- [x] Unit tests for `getErc20Balance` (success, RPC error, network failure)
- [x] Unit tests for `getNativeBalance` (success, RPC error, network failure)
- [x] Unit tests for `resolveWrappedNativeToken` (non-wrapped, sufficient WETH, zero WETH + native ETH, both insufficient, case-insensitive, solana, null inputs)
- [x] Integration tests for quote handler (auto-substitution, no substitution with WETH balance, no substitution for sentinel)
- [x] All 621 tests pass

### DIFF
--- a/src/__tests__/trading.test.js
+++ b/src/__tests__/trading.test.js
@@ -753,121 +753,71 @@ describe('stripLeadingZeros', () => {
 
 // ============= Balance Helpers =============
 
-describe('getErc20Balance', () => {
-  it('should return balance from mocked RPC response', async () => {
-    const origFetch = global.fetch;
-    // 1000000000000000000 = 0xde0b6b3a7640000
-    global.fetch = vi.fn().mockResolvedValue({
+describe('balance helpers', () => {
+  afterEach(() => vi.restoreAllMocks());
+
+  it('getErc20Balance should return parsed balance from RPC', async () => {
+    vi.spyOn(global, 'fetch').mockResolvedValue({
       ok: true,
       json: async () => ({ jsonrpc: '2.0', id: 1, result: '0x0000000000000000000000000000000000000000000000000de0b6b3a7640000' }),
     });
-
     const balance = await getErc20Balance('base', '0x4200000000000000000000000000000000000006', '0x' + 'ab'.repeat(20));
     expect(balance).toBe(1000000000000000000n);
-
-    global.fetch = origFetch;
   });
 
-  it('should return 0n on RPC error', async () => {
-    const origFetch = global.fetch;
-    global.fetch = vi.fn().mockResolvedValue({
-      ok: true,
-      json: async () => ({ jsonrpc: '2.0', id: 1, error: { code: -32000, message: 'execution reverted' } }),
-    });
-
-    const balance = await getErc20Balance('base', '0x4200000000000000000000000000000000000006', '0x' + 'ab'.repeat(20));
-    expect(balance).toBe(0n);
-
-    global.fetch = origFetch;
-  });
-
-  it('should return 0n on network failure', async () => {
-    const origFetch = global.fetch;
-    global.fetch = vi.fn().mockRejectedValue(new Error('network error'));
-
-    const balance = await getErc20Balance('base', '0x4200000000000000000000000000000000000006', '0x' + 'ab'.repeat(20));
-    expect(balance).toBe(0n);
-
-    global.fetch = origFetch;
-  });
-});
-
-describe('getNativeBalance', () => {
-  it('should return balance from mocked RPC response', async () => {
-    const origFetch = global.fetch;
-    global.fetch = vi.fn().mockResolvedValue({
+  it('getNativeBalance should return parsed balance from RPC', async () => {
+    vi.spyOn(global, 'fetch').mockResolvedValue({
       ok: true,
       json: async () => ({ jsonrpc: '2.0', id: 1, result: '0xde0b6b3a7640000' }),
     });
-
     const balance = await getNativeBalance('base', '0x' + 'ab'.repeat(20));
     expect(balance).toBe(1000000000000000000n);
-
-    global.fetch = origFetch;
   });
 
-  it('should return 0n on RPC error', async () => {
-    const origFetch = global.fetch;
-    global.fetch = vi.fn().mockResolvedValue({
-      ok: true,
-      json: async () => ({ jsonrpc: '2.0', id: 1, error: { code: -32000, message: 'error' } }),
-    });
-
-    const balance = await getNativeBalance('base', '0x' + 'ab'.repeat(20));
-    expect(balance).toBe(0n);
-
-    global.fetch = origFetch;
-  });
-
-  it('should return 0n on network failure', async () => {
-    const origFetch = global.fetch;
-    global.fetch = vi.fn().mockRejectedValue(new Error('network error'));
-
-    const balance = await getNativeBalance('base', '0x' + 'ab'.repeat(20));
-    expect(balance).toBe(0n);
-
-    global.fetch = origFetch;
+  it('both helpers should return 0n on failure', async () => {
+    vi.spyOn(global, 'fetch').mockRejectedValue(new Error('network error'));
+    expect(await getErc20Balance('base', '0x4200000000000000000000000000000000000006', '0x' + 'ab'.repeat(20))).toBe(0n);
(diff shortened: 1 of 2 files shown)


src/__tests__/trading.test.js
```diff
@@ -25,6 +25,9 @@ import {
   buildApprovalTransaction,
   stripLeadingZeros,
   buildTradingCommands,
+  getErc20Balance,
+  getNativeBalance,
+  resolveWrappedNativeToken,
```
Codier, line 30: 🟡 **Should Fix — Test bloat**: `getErc20Balance` and `getNativeBalance` have near-identical test suites (success, RPC error, network failure) — 6 tests that are essentially copy-pasted with different method names. These are thin wrappers around `fetch` → `BigInt()` with a catch-all `0n`. The error/network-failure cases all test the same `catch { return 0n }` path.

Suggest: collapse to **1 test each** (happy path) and **1 shared test** for the error fallback behavior, or just test them implicitly through `resolveWrappedNativeToken` which exercises both. That cuts ~40 lines of near-duplicate mocking.

src/__tests__/trading.test.js
```diff
@@ -25,6 +25,9 @@ import {
...
 } from '../trading.js';
 import { keccak256, rlpEncode } from '../crypto.js';
```
Codier, line 32: 🟡 **Should Fix**: Every test manually saves/restores `global.fetch` with `origFetch`. This is fragile — if an assertion throws before the restore, `global.fetch` stays mocked for subsequent tests. Use `vi.spyOn(global, "fetch")` with `afterEach(() => vi.restoreAllMocks())` instead. That's 6+ places where this pattern repeats.

src/__tests__/trading.test.js
```diff
@@ -748,6 +751,335 @@ describe('stripLeadingZeros', () => {
...
+      if (callCount === 1) {
+        // getErc20Balance: zero WETH
+        return { ok: true, json: async () => ({ jsonrpc: '2.0', id: 1, result: '0x0' }) };
+      }
+      // getNativeBalance: 1 ETH
+      return { ok: true, json: async () => ({ jsonrpc: '2.0', id: 1, result: '0xde0b6b3a7640000' }) };
+    });
+
+    const result = await resolveWrappedNativeToken(
+      '0x4200000000000000000000000000000000000006', 'base', '0x' + 'ab'.repeat(20), '10000000000000000'
```
Codier, line 822: 🟢 **Nitpick — Low-value test**: The `null/undefined inputs` test just verifies early returns from null guards. These are trivially obvious from reading the code and don't test any real user scenario. Same for the Solana test — it just hits the "no entry in WRAPPED_NATIVE_TOKENS" early return. Consider removing or combining into a single "unsupported inputs" case.

src/__tests__/trading.test.js
```diff
@@ -748,6 +751,335 @@ describe('stripLeadingZeros', () => {
...
+    );
+    expect(result.substituted).toBe(true);
+    expect(result.address).toBe('0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee');
+    expect(result.symbol).toBe('WETH');
+    expect(result.nativeSymbol).toBe('ETH');
+
+    global.fetch = origFetch;
+  });
```
Codier, line 828: 🟡 **Should Fix — Test size**: The 3 integration tests for the quote handler are ~150 lines of mock setup that are mostly identical (same mock structure, same quote response, same wallet creation). Extract a shared helper like `function mockQuoteFetch({ erc20Balance, nativeBalance, ... })` and each test shrinks to ~15 lines. As-is these are hard to review because the actual assertion is buried under walls of mock boilerplate.

src/trading.js
```diff
@@ -643,7 +651,96 @@ async function promptPassword(prompt, deps = {}) {
...
+    return BigInt(body.result);
+  } catch {
+    return 0n;
+  }
+}
+
+/**
+ * Auto-detect whether to substitute a wrapped native token (WETH/WBNB) with the native sentinel.
+ * Checks on-chain balances: if user lacks wrapped token but has native token, substitutes.
+ *
```
Codier, line 717: 🔴 **Blocking — Correctness**: `resolveWrappedNativeToken` calls `BigInt(amount)` with no guard. If `amount` is empty string, undefined, or a non-numeric string, this throws a `SyntaxError` that isn't caught. Since this is called with user-provided CLI input (`options.amount`), it needs validation or a try/catch. The "neither sufficient" fallback would be the right behavior for invalid amounts.

src/trading.js
```diff
@@ -23,6 +23,14 @@ const CHAIN_MAP = {
   bsc:      { index: '56',  type: 'evm',    chainId: 56,   name: 'BSC',      explorer: 'https://bscscan.com/tx/' },
 };
 
+const NATIVE_TOKEN_SENTINEL = '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee';
```
Codier, line 26: 🟢 **Nitpick**: The old regex `/^0x[eE]{40}$/` accepted mixed-case `0xEeEe...`, while the new code lowercases before comparing. Functionally equivalent, but this is an unrelated refactor — consider keeping it out of this PR for cleaner history, or at least noting it in the PR description.

Codier (commented):
1 🔴 blocking (uncaught BigInt throw on invalid amount input), 3 🟡 should-fix (test bloat/duplication, fragile fetch mocking, integration tests need helper extraction), 2 🟢 nitpicks. Effort 2/4.

src/trading.js
Codier, line 26: 🟢 **Nitpick**: `isNativeToken` (line 649) still uses the old regex `/^0x[eE]{40}$/` while the new code uses the `NATIVE_TOKEN_SENTINEL` constant. Minor inconsistency — could update `isNativeToken` to use the constant too for a single source of truth. Not blocking.

Codier (commented):
LGTM — all previous feedback addressed. Tests are clean now (proper vi.spyOn + restoreAllMocks, shared mockQuoteFetch helper, consolidated edge cases). BigInt guard added. One cosmetic nitpick remaining. Effort 1/4.

Codier:
## Manual Verification (live RPC on Base)

Tested `resolveWrappedNativeToken` against real on-chain balances using `--from 0x4200000000000000000000000000000000000006` (WETH on Base) with `--amount 1000000000000000` (0.001 ETH):

| Address | Native ETH | WETH | Result |
|---|---|---|---|
| `0x7236...` (local wallet) | 0 | 0 | No substitution — neither sufficient, API handles error |
| `0xd8dA...` (vitalik.eth) | ~0.071 ETH | ~0.063 ETH | No substitution — WETH balance >= amount |
| `0x388C...` (random EOA) | ~0.0015 ETH | 0 | **Substituted** `--from` to `0xeee...eee` native sentinel |

All three code paths confirmed working against live Base RPC (`https://mainnet.base.org`).

PR #93, opened 2026-02-26
fix: warn when --from is wrapped native token or native sentinel

### LINKED ISSUE #80: Bug: WETH address as --from causes TRANSFER_FROM_FAILED when user has native ETH

Codier:
## Summary

- Adds a **stderr warning** when `--from` is a wrapped native token (WETH/WBNB) or the native sentinel (`0xeee...eee`), so AI agents can detect the mismatch before execution fails with `TRANSFER_FROM_FAILED`
- Covers both directions: wrapped → suggests native sentinel, native sentinel → suggests wrapped address
- No behavior change — the quote proceeds exactly as requested; the warning is informational only
- Replaces the reverted auto-substitution approach (PR #85) with a non-invasive warning

## Why

On Base, using the WETH address (`0x4200...0006`) as `--from` causes `TRANSFER_FROM_FAILED` at execution time when the wallet holds native ETH, not WETH. The previous fix (PR #85) silently substituted the token, but was reverted because the CLI shouldn't change what the user requested. This PR warns instead, giving the calling agent actionable info to self-correct.

## Changes

- **`src/trading.js`**: `WRAPPED_NATIVE_TOKENS` constant + exported `getWrappedNativeFromWarning()` pure function + 2-line call in quote handler
- **`src/__tests__/trading.test.js`**: 9 unit tests (WETH/WBNB on all supported chains, both directions, case-insensitivity, null safety)

## Test plan

- [x] `npm test` — all 617 tests pass
- [ ] Manual: `nansen trade quote --chain base --from 0x4200000000000000000000000000000000000006 --to 0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913 --amount 10000000000000000` — shows warning suggesting native sentinel
- [ ] Manual: `nansen trade quote --chain base --from 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee --to 0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913 --amount 10000000000000000` — shows warning suggesting WETH address

Closes #80

### DIFF
--- a/src/trading.js
+++ b/src/trading.js
@@ -23,6 +23,7 @@ const CHAIN_MAP = {
   bsc:      { index: '56',  type: 'evm',    chainId: 56,   name: 'BSC',      explorer: 'https://bscscan.com/tx/' },
 };
 
+// Extend when adding new EVM chains (e.g. arbitrum WETH, polygon WMATIC)
 const WRAPPED_NATIVE_TOKENS = {
   ethereum: { address: '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2', symbol: 'WETH', nativeSymbol: 'ETH' },
   base:     { address: '0x4200000000000000000000000000000000000006', symbol: 'WETH', nativeSymbol: 'ETH' },
@@ -670,7 +671,7 @@ export function getWrappedNativeFromWarning(tokenAddress, chain) {
   }
 
   // Case 2: --from is native sentinel — mention the wrapped alternative
-  if (/^0x[eE]{40}$/.test(tokenAddress)) {
+  if (isNativeToken(tokenAddress)) {
     return `Warning: --from is native ${wrapped.nativeSymbol}. ` +
       `If you hold ${wrapped.symbol} instead, use: ${wrapped.address}`;
   }


askeluv (dismissed):
Clean PR — looks good to merge. A few minor observations (non-blocking):

1. **Reuse `isNativeToken()`**: The regex `/^0x[eE]{40}$/` on L654 already exists as a function — `getWrappedNativeFromWarning` duplicates it inline. Trivial, but could reuse it.

2. **Future chains**: Arbitrum and Polygon also have wrapped native tokens (WETH, WMATIC). Worth a comment noting `WRAPPED_NATIVE_TOKENS` should be extended when those chains are added.

Otherwise solid — pure function, good test coverage, stderr-only (won't break piped output), and the non-invasive approach is the right call after the PR #85 revert. 👍

PR #328, opened 2026-03-23
feat: add Jupiter Limit Order V2 support (Solana)

Pepewitch:
## Summary

Add `trade limit` commands for Jupiter Limit Order V2 on Solana — create, list, cancel, and update trigger-based orders.

### Commands
- `trade limit create` — place a limit order (deposit + order in one flow)
- `trade limit list` — list orders with `--state active|past` filtering, pagination, mint filter
- `trade limit cancel` — cancel an order (withdrawal + confirmation)
- `trade limit update` — update trigger price and/or slippage on an existing order

### Design decisions
- **Amount convention matches swap**: base units by default, `--amount-unit token` for human-readable (e.g. `--amount 1.5 --amount-unit token`)
- **Slippage convention matches swap**: `--slippage 0.03` (decimal, 3%) — converted internally to bps for the API
- **Shared utilities from trading.js**: `resolveTokenDecimals()`, `convertToBaseUnits()`, `validateBaseUnitAmount()`, `resolveTokenAddress()` — no duplicate infrastructure
- **JWT auth**: challenge-response with 23h TTL disk cache at `~/.nansen/limit-order-auth.json`
- **Wallet support**: local wallets, Privy, and WalletConnect
- **Auto vault registration**: first-time users get a vault registered transparently
- **5xx ghost-order recovery**: if `createOrder` returns 5xx, checks recent orders (matching token pair, amount, trigger price, and timestamp) to warn about duplicates before the user retries

### Files
| File | What |
|------|------|
| `src/limit-order.js` | Core module: auth, signing, wallet resolution, all 4 command handlers |
| `src/cli.js` | Route `trade limit <sub>` to limit-order handlers |
| `src/schema.json` | Option docs for create, list, cancel, update |
| `src/privy.js` | `signSolanaMessage()` for Privy wallets |
| `src/walletconnect-trading.js` | `signSolanaMessageViaWalletConnect()` for WC wallets |
| `src/__tests__/limit-order.test.js` | 96 unit tests |

## Test plan

- [x] `npm test` — 1378 tests pass (96 new for limit-order)
- [x] `npm run lint` — clean
- [x] Dogfooded locally:
  - Created orders: SOL to USDC, SOL to TRUMP, SOL to ORCA (all succeeded)
  - Created order: SOL to PUMP (correctly rejected — transfer hook extension)
  - Listed orders with `--state active`, verified formatting
  - Updated order (trigger price + slippage), verified via list
  - Cancelled all orders, verified withdrawal transactions on Solscan
  - Insufficient balance returns clean error (no misleading ghost-order warning)
- [x] 5xx ghost-order check: only triggers after `createOrder` call, not on pre-submission errors
- [x] Schema.json updated with correct API state values (`active`/`past`)

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -1487,7 +1487,7 @@ SUBCOMMANDS:
   quote          Get a swap quote (price, route, fees)
   execute        Sign and broadcast a quoted swap
   bridge-status  Check cross-chain bridge transaction status
-  limit-order    Limit order management (Solana only, Jupiter Trigger V2)
+  limit-order    Limit order management (Solana only)
 
 USAGE:
   nansen trade quote --chain <chain> --from <token> --to <token> --amount <units> [--wallet <name>]
@@ -1527,7 +1527,7 @@ CROSS-CHAIN NOTES (when using --to-chain):
     if (sub === 'limit-order') {
       const loSub = args[1];
       if (!loSub || loSub === 'help') {
-        log(`nansen trade limit-order — Limit order commands (Solana only, Jupiter Trigger V2)
+        log(`nansen trade limit-order — Limit order commands (Solana only)
 
 SUBCOMMANDS:
   create    Place a new limit order
--- a/src/cli.js
+++ b/src/cli.js
@@ -1537,7 +1537,7 @@ SUBCOMMANDS:
 
 USAGE:
   nansen trade limit-order create --from <token> --to <token> --amount <units> --trigger-mint <token> --trigger-condition <above|below> --trigger-price <usd>
-  nansen trade limit-order list [--state <open|filled|cancelled|expired>]
+  nansen trade limit-order list [--state <open|filled|cancelled|expired|all>]
   nansen trade limit-order cancel --order <orderId>
   nansen trade limit-order update --order <orderId> --trigger-price <usd>`);
         return;
--- a/src/cli.js
+++ b/src/cli.js
@@ -6,6 +6,7 @@
 import { NansenAPI, NansenError, CommandError, ErrorCode, saveConfig, deleteConfig, getConfigFile, clearCache, getCacheDir, validateAddress, normalizeAddress, sleep } from './api.js';
 import { buildWalletCommands } from './wallet.js';
 import { buildTradingCommands } from './trading.js';
+import { buildLimitOrderCommands } from './limit-order.js';
 import { formatAlertsTable, buildAlertsCommands } from './commands/alerts.js';
 import { buildAgentCommands } from './commands/agent.js';
 import { resolveAddress, isEnsName } from './ens.js';
@@ -1474,8 +1475,9 @@ export function buildCommands(deps = {}) {
     return cmds[category](args.slice(1), apiInstance, flags, options);
   };
 
-  // 'trade' delegates to quote/execute from buildTradingCommands
+  // 'trade' delegates to quote/execute from buildTradingCommands and limit-order from buildLimitOrderCommands
   const tradingCmds = buildTradingCommands(deps);
+  const limitOrderCmds = buildLimitOrderCommands(deps);
   cmds['trade'] = async (args, apiInstance, flags, options) => {
     const sub = args[0];
     if (!sub || sub === 'help') {
@@ -1485,12 +1487,14 @@ SUBCOMMANDS:
   quote          Get a swap quote (price, route, fees)
   execute        Sign and broadcast a quoted swap
   bridge-status  Check cross-chain bridge transaction status
+  limit-order    Limit order management (Solana only, Jupiter Trigger V2)
 
 USAGE:
   nansen trade quote --chain <chain> --from <token> --to <token> --amount <units> [--wallet <name>]
   nansen trade quote --chain <chain> --to-chain <chain> --from <token> --to <token> --amount <units>
   nansen trade execute --quote <quoteId> [--wallet <name>]
   nansen trade bridge-status --tx-hash <hash> --from-chain <chain> --to-chain <chain>
+  nansen trade limit-order <create|list|cancel|update> [options]
 
 EXAMPLES:
   nansen trade quote --chain solana --from SOL --to USDC --amount 1000000000
@@ -1498,9 +1502,11 @@ EXAMPLES:
   nansen trade quote --chain base --to-chain solana --from USDC --to USDC --amount 1000000
   nansen trade execute --quote 1708900000000-abc123
   nansen trade bridge-status --tx-hash 0xabc... --from-chain base --to-chain solana
+  nansen trade limit-order create --from SOL --to USDC --amount 1000000000 --trigger-price 80
+  nansen trade limit-order list
 
 WALLET:
-  --wallet <name>   Use a named wallet, or "walletconnect" / "wc" for WalletConnect (EVM only).
+  --wallet <name>   Use a named wallet, or "walletconnect" / "wc" for WalletConnect.
                     Defaults to the default local wallet if omitted.
 
 SYMBOLS:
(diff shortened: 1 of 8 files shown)


src/cli.js
```diff
@@ -1430,38 +1431,65 @@ export function buildCommands(deps = {}) {
...
+        log(`nansen trade limit-order — Limit order commands (Solana only)
+
+SUBCOMMANDS:
+  create    Place a new limit order
+  list      List your limit orders
+  cancel    Cancel an open order
+  update    Update trigger price or slippage
+
+USAGE:
+  nansen trade limit-order create --from <token> --to <token> --amount <units> --trigger-mint <token> --trigger-condition <above|below> --trigger-price <usd>
```
TimNooren, line 1539: IIUC the `unit` here is the human-readable amount. In `nansen trade quote/execute` we use raw amounts (i.e. not adjusting for decimals). Maybe we should be consistent at least? We can add a `--amount-unit`, similar to https://github.com/nansen-ai/nansen-cli/pull/324.

Also, I saw that this needs to be >$10. Might also be good to document.

src/schema.json
```diff
@@ -786,6 +786,138 @@
...
+                  "description": "Trigger when price is 'above' or 'below' target"
+                },
+                "trigger-mint": {
+                  "type": "string",
+                  "required": true,
+                  "description": "Token whose price triggers the order (e.g. SOL)"
+                },
+                "slippage-bps": {
+                  "type": "number",
+                  "description": "Slippage tolerance in basis points (e.g. 50 = 0.5%)"
```
TimNooren, line 948: In `nansen trade quote/execute` we use `slippage` and it's a human-readable percentage. Would be good to be consistent.

src/schema.json
```diff
@@ -786,6 +786,138 @@
...
+                },
+                "to": {
+                  "type": "string",
+                  "required": true,
+                  "description": "Token to buy (address or symbol)"
+                },
+                "amount": {
+                  "type": "string",
+                  "required": true,
+                  "description": "Amount in base units (e.g. lamports)"
```
TimNooren, line 929: This is currently not correct right?

src/limit-order.js
```diff
@@ -0,0 +1,868 @@
...
+  saveCachedToken(walletPubkey, token);
+
+  return token;
+}
+
+// ============= Wallet Resolution =============
+
+/**
+ * Resolve a Solana wallet for limit orders.
+ * Follows the same 3-way dispatch as trading.js: WalletConnect / named / default.
```
TimNooren, line 255: Can we extract this code then instead of duplicating?

src/limit-order.js
```diff
@@ -0,0 +1,868 @@
...
+
+    'update': async (args, apiInstance, flags, options) => {
+      const orderId = options.order || options['order-id'] || args[0];
+      const triggerPrice = options['trigger-price'];
+      const slippageBps = options['slippage-bps'];
+      const walletName = options.wallet;
+
+      if (!orderId) {
+        log(`
+Usage: nansen trade limit-order update --order <orderId> [--trigger-price <usd>] [--slippage-bps <bps>]
```
TimNooren, line 801: 💭 `nansen trade limit`?

TimNooren (commented):
@Pepewitch Thanks🙌 Some small comments. My main concern is the interface, since it's going to be difficult to change once released. I think the current `nansen trade quote/execute` interface is not ideal, but I'd say having at least a consistent interface for now is more important, so let's stick with it.

Let's also add a changeset file🙏

src/limit-order.js
```diff
@@ -0,0 +1,868 @@
...
+
+  const headers = {
+    'Accept': 'application/json',
+    'Content-Type': 'application/json',
+  };
+  if (token) {
+    headers['Authorization'] = `Bearer ${token}`;
+  }
+  if (process.env.NANSEN_API_KEY) {
+    headers['X-API-Key'] = process.env.NANSEN_API_KEY;
```
TimNooren, line 98: This should probably not be included.

src/limit-order.js
```diff
@@ -0,0 +1,868 @@
...
+    'create': async (args, apiInstance, flags, options) => {
+      const fromRaw = options.from || options['from-token'] || args[0];
+      const toRaw = options.to || options['to-token'] || args[1];
+      const from = resolveTokenAddress(fromRaw, 'solana');
+      const to = resolveTokenAddress(toRaw, 'solana');
+      const amount = options.amount || args[2];
+      const triggerPrice = options['trigger-price'];
+      const triggerCondition = options['trigger-condition'];
+      const triggerMintRaw = options['trigger-mint'];
+      const slippageBps = options['slippage-bps'] != null ? Number(options['slippage-bps']) : undefined;
```
0xlaveen, line 487: `create` doesn't validate `slippageBps` range client-side, but `update` does (line 831). If a user passes `--slippage-bps 20000`, it passes through to the API which rejects it with: `Too big: expected number to be <=10000`. The error message is decent, but for consistency with `update` we should add the same client-side check here:

```js
if (slippageBps != null && (slippageBps < 0 || slippageBps > 10000)) {
  log('Error: --slippage-bps must be between 0 and 10000 basis points.');
  exit(1);
  return;
}
```

src/limit-order.js
```diff
@@ -0,0 +1,868 @@
...
+          inputAmount: amountBaseUnits,
+          outputMint: to,
+          triggerMint,
+          triggerCondition,
+          triggerPriceUsd: price, // Must be Number, not string
+          ...(slippageBps != null ? { slippageBps } : {}),
+          ...(expiresAt != null ? { expiresAt } : {}),
+        };
+
+        const result = await createOrder(token, orderParams);
```
0xlaveen, line 656: This is the most impactful UX issue I found during testing. When `createOrder` returns a 500 Gateway Timeout (which happened intermittently from Jupiter), the CLI reports an error — but the order was actually created on the backend. The user thinks it failed, retries, and creates duplicates. During testing I accumulated 7 ghost orders this way.

Suggestion: catch timeout/5xx errors specifically and follow up with a `listOrders` check:

```js
} catch (err) {
  if (err.status >= 500) {
    log(`Warning: Order submission returned an error, but may have succeeded.`);
    log(`  Checking order status...`);
    try {
      const check = await listOrders(token, pubkey, { limit: 1 });
      const recent = check.orders?.[0];
      if (recent && recent.inputMint === from && recent.outputMint === to) {
        log(`  Found recent order ${recent.id} (status: ${recent.status}). It may be yours.`);
        log(`  Run: nansen trade limit-order list`);
        return;
      }
    } catch { /* list check failed, fall through to error */ }
  }
  log(`Error: ${err.message}`);
  // ...
}
```

This saved me a lot of confusion during testing — orders were silently created despite the error output.

0xlaveen (commented):
## Smoke test results (Privy wallet, Solana mainnet)

Ran 42 end-to-end tests against a Privy server wallet (`Hm6a7ZkpPqTHZVqueaPNQRD4t7mg9DKPECpGo1QcFXjR`) on mainnet. Full lifecycle verified across SOL→USDC, SOL→JUP, and SOL→WBTC (Wormhole) pairs. All 4 commands work. All 1180 unit tests pass.

**Bug fixed during testing:** `--state` query param should be `--status` — the API rejects `state` with "Request validation failed". Fix is in a follow-up commit on the branch.

Two issues flagged below in inline comments.

— Ari

src/limit-order.js
TimNooren: Should trading-api protect against this instead of the user? And we should maybe rely on `depositRequestId` instead of  `ecent.inputMint === from && recent.outputMint === to`? @Pepewitch

src/limit-order.js
```diff
@@ -0,0 +1,904 @@
...
+  return `${humanAmount} ${info.symbol} (${amount} base units)`;
+}
+
+function formatTimestamp(ts) {
+  if (!ts) return '?';
+  const num = Number(ts);
+  if (isNaN(num)) return ts;
+  const date = new Date(num);
+  return `${date.toLocaleString()} (${date.toISOString()})`;
+}
```
kome12, line 447: Small comment: This handles date differently for numeric and string timestamps I think. I didn't come across it while testing so I'l leave it to you as you probably know better.

```suggestion
function formatTimestamp(ts) {
    if (!ts) return '?';
    const date = new Date(ts); // handles both ISO strings and epoch ms
    if (isNaN(date.getTime())) return String(ts);
    return `${date.toLocaleString()} (${date.toISOString()})`;
  }
```

0xlaveen:
## Dogfood Results

Full round-trip tested: **list → create → verify → cancel → verify gone** ✅

| Step | Result |
|------|--------|
| `limit-order list` (empty wallet) | "No limit orders found" ✅ |
| `limit-order create --amount 0.15 SOL → USDC, trigger >$999` | Order created, deposit tx on-chain ✅ |
| `limit-order list --state active` | Shows 1 Open order ✅ |
| `limit-order cancel --order <id>` | Cancelled, withdrawal tx on-chain ✅ |
| `limit-order list --state active` | Empty again ✅ |
| Balance: 0.2 SOL → 0.1938 SOL | ~0.006 SOL gas for 2 txs ✅ |

### Bugs found & fixed (last commit)

1. **Vault field mismatch** — API returns `vaultAddress`, code checked `vaultPubkey`. Vault check always failed → re-registration → hard error.
2. **"Vault already registered" not caught** — When #1 triggered re-registration, it threw instead of continuing.
3. **`--state` allowed values wrong** — Help/schema said `open|filled|cancelled|expired`, API only accepts `active|past`.

0xlaveen (approved):
Dogfooded full round-trip (create → list → cancel). All flows work. Three bugs found and fixed in the last commit.

PR #409, opened 2026-04-14
feat: add top-tokens subcommand for Nansen Score discovery

jake-kennis:
## Summary
- Add `nansen research token top-tokens` subcommand that calls `POST /api/internal/nansen-score-top-tokens`
- Supports `--market-cap` filter (lowcap/midcap/largecap) and `--limit`
- Fills the discovery gap: users can now find top-scoring tokens, then drill into individual indicators

## Test plan
- [x] All 1323 existing tests pass
- [x] 3 new unit tests for `topTokens()` API method (endpoint, marketCapGroup passthrough, default limit)
- [x] CLI integration test asserts `top-tokens` in help output
- [x] Coverage test updated
- [x] Schema validated via `nansen schema token`

### DIFF
--- a/.changeset/top-tokens-nansen-score.md
+++ b/.changeset/top-tokens-nansen-score.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": minor
+---
+
+Add `top-tokens` subcommand to discover top-scoring tokens by Nansen Score. Calls internal endpoint (`/api/internal/nansen-score-top-tokens`) with optional `--market-cap` filter.
--- a/skills/nansen-token-screener/SKILL.md
+++ b/skills/nansen-token-screener/SKILL.md
@@ -45,10 +45,18 @@ nansen research token indicators --token $TOKEN --chain $CHAIN
 # Flow intelligence — only use for promising tokens from screener/indicators above
 nansen research token flow-intelligence --token $TOKEN --chain $CHAIN
 # → net_flow_usd per label: smart_trader, whale, exchange, fresh_wallets, public_figure
+
+# Nansen Score Top Tokens — "what should I buy?" (internal, @nansen.ai only)
+# Use this FIRST for discovery, then drill into individual tokens with `indicators` above
+nansen research token top-tokens --limit 25
+nansen research token top-tokens --market-cap largecap --limit 10
+# → token_symbol, chain, performance_score, risk_score, plus per-indicator contribution fields
 ```
 
 Screener timeframes: `5m`, `10m`, `1h`, `6h`, `24h`, `7d`, `30d`
 
 Indicators: score is "bullish"/"bearish"/"neutral". signal_percentile > 70 = historically significant. Some tokens return empty indicators — not an error.
+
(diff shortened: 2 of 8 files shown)


jake-kennis:
Howdy @marius-reed I tried following what you did for prediction markets, touching the same files and such. Ready for your eyes whenever. Thank you!

0xlaveen (changes requested):
## Bug: `--market-cap` filter sends camelCase field name, API expects snake_case

Tests all pass (1323/1323) and the basic `top-tokens` command works. However, the `--market-cap` filter is broken in production:

```
$ nansen research token top-tokens --market-cap largecap --limit 3
Field 'marketCapGroup' is not recognized. This API uses snake_case field names - did you mean 'market_cap_group'?
```

### Fix needed in `src/api.js:1209`

```diff
-    if (marketCapGroup) body.marketCapGroup = marketCapGroup;
+    if (marketCapGroup) body.market_cap_group = marketCapGroup;
```

The unit test for `marketCapGroup` passthrough (`api.test.js`) asserts the camelCase key, so it passes — but it doesn't match what the API actually expects. The test should also be updated to assert `market_cap_group`.

Everything else looks good — clean implementation, proper changeset, schema/SKILL.md updates, coverage test updated.

0xlaveen:
## Missing: field definitions for top-tokens response

The SKILL.md and schema don't explain what any of the response fields mean. The agent consuming this has no way to interpret the data beyond two vague heuristics (`performance_score >= 15 = buy candidate`, `risk_score > 0 = safer`).

Fields that need definitions:

| Field | Question |
|---|---|
| `performance_score` | Scale/range? How is it composed from the sub-indicators below? |
| `risk_score` | Can go negative (e.g. FET = -40). Does negative = riskier? What's the range? |
| `price_momentum_performance` | What does 30 mean? Out of what? |
| `trading_range_performance` | Only non-zero for ETH/WBTC (15). What triggers this? |
| `chain_fees_performance` | Zero for every token in our test. When is this non-zero? |
| `chain_tvl_performance` | Same — always 0. |
| `protocol_fees_performance` | Same — always 0. |
| `btc_reflexivity_risk` | Ranges from -20 to 40. What does this measure? |
| `liquidity_risk` | Goes negative (-20). Negative = more risky or less risky? |
| `concentration_risk` | When does this fire? Only saw 0 or 10. |
| `inflation_risk` | Same — 0 or 10. Threshold? |

Without these, the LLM will hallucinate interpretations. Even a one-liner per field in the SKILL.md would go a long way — e.g.:

```
# Field definitions (all scores are additive components, can be negative)
# performance_score: sum of *_performance fields. Higher = stronger bullish signal. Range: ~ -100 to 100.
# risk_score: sum of *_risk fields. Positive = safer, negative = riskier. Range: ~ -100 to 100.
# price_momentum_performance: contribution from recent price trend. 30 = strong uptrend.
# btc_reflexivity_risk: correlation risk to BTC. Positive = moves with BTC (safer), negative = decoupled.
# liquidity_risk: based on DEX/CEX depth. Positive = liquid, negative = thin books.
...
```

(Guessing at the above — would need the actual definitions from whoever built the scoring model.)

PR #422, opened 2026-04-27
docs+feat: MPP/Tempo guidance + x402 support for USDT0 on X Layer

gulshngill:
## Summary

Two related additions for paid-access auth, plus the cleanup that fell out of testing them end-to-end.

1. **MPP / Tempo guidance** — document MPP (Tempo's stablecoin rail) as a third paid-access option alongside API key and x402. `nansen-cli` does **not** sign MPP credentials; users install [tempo](https://docs.tempo.xyz) separately and call the API via `tempo request`. Everything here is documentation, not new payment code.
2. **x402 support for USDT0 on X Layer** — adds X Layer (`eip155:196`) USDT0 alongside Base USDC and Solana SPL USDC. EIP-3009 signing already worked generically off the 402 `accepts` list (network/asset/EIP-712 name+version are all read from the server response), so the only client changes are an X Layer RPC entry and a network-aware balance-check lookup.
3. **Balance-warning fix** — `checkX402Balance()` now returns `{ balance, symbol }` so the low-balance warning reports the actual stablecoin (USDC vs. USDT0) instead of always saying "USDC".

## Files changed

**Docs**
- `README.md` — Auth section restructured into 3 numbered options; new "MPP / Tempo" section; rail-comparison table extended with X Layer.
- `AGENTS.md` — Paid-Access-Rails section updated to mention X Layer USDT0 + the generic-from-402 signing approach.
- `skills/nansen-mpp-payment/SKILL.md` — New skill covering tempo install / `tempo request` examples / x402-vs-MPP comparison.

**Code**
- `src/api.js` — No-API-key 402 error lists three options (API key, x402 with Base/Solana/X-Layer, MPP via tempo). `_x402Retry` low-balance warning uses the dynamic stablecoin symbol from `checkX402Balance()`.
- `src/rpc-urls.js` — New `xlayer` entry with `NANSEN_XLAYER_RPC` env-var override.
- `src/x402.js` — `checkX402Balance()` now picks token + RPC + symbol from a per-network table keyed on the requirement's `network` field, defaulting to Base USDC for unknown EVM networks. Returns `{ balance, symbol }`.

**Changesets**
- `.changeset/mpp-tempo-guidance.md` — patch (docs).
- `.changeset/x402-xlayer-usdt0.md` — minor (new payment network).
- `.changeset/x402-balance-warning-and-xlayer-notes.md` — patch (dynamic symbol fix).

## Notes for reviewer

- Wrong USDT0 contract address from the first round of this PR (`0x779DED2B…7736`) was caught in review and replaced with the correct `0x779Ded0c9e1022225f8E0630b35a9b54bE713736`. The signing path was always correct (it reads `requirements.asset` from the API), but the `checkX402Balance()` RPC call would have queried a phantom address. Test coverage of the new `EVM_NETWORKS['eip155:196']` path is a reasonable follow-up.
- The X Layer rail was tested end-to-end against dev and is now signing/settling cleanly. (The original test surfaced an upstream EIP-712 domain-hash encoding bug; that has since been fixed and shipped, and the temporary "unavailable" warnings that lived on this branch for a day have been removed.)

## Test plan

- [x] `npm run lint` passes
- [x] `npm test` passes (1394 passed, 2 skipped)
- [x] Manual: unauthenticated request with no wallet → 402 error lists three options
- [x] Manual: x402 round-trip on X Layer USDT0 against dev — verify + settle succeed; `Payment-Receipt` returned
- [x] Manual: x402 round-trip on Base USDC against dev — unaffected (regression check)
- [ ] Manual: skim README "MPP / Tempo" section for accuracy of `tempo wallet` / `tempo request` examples against the actual Tempo CLI surface


0xlaveen:
**CLI bug: wrong USDT0 contract address on X Layer**

The hardcoded X Layer USDT0 in `src/x402.js:195` `EVM_NETWORKS` is wrong:

- PR: \`0x779DED2B30f6f45db42e0aB1130fC1b2A0a37736\` — no contract at this address (`eth_getCode` → empty)
- Real USDT0 on X Layer: \`0x779Ded0c9e1022225f8E0630b35a9b54bE713736\` — `symbol()` returns `USD₮0`, and this is also the address the prod API itself advertises in `Payment-Required.accepts` for `eip155:196`

Impact is limited because signing reads `requirements.asset` from the API response (so payments still sign correctly against the right contract), but `checkX402Balance('eip155:196')` queries the phantom address → result is `NaN` → the "low balance" warning silently never fires for X Layer wallets.

Suggested fix: replace the literal in `EVM_NETWORKS['eip155:196'].token` with `0x779Ded0c9e1022225f8E0630b35a9b54bE713736`. Optional: drive `EVM_NETWORKS` from the requirement itself instead of hardcoding, since the API is already authoritative on `asset`.

0xlaveen:
**Server-side bug discovered while testing the X Layer rail end-to-end**

Not a PR-blocker, but flagging here because the README/skill added in this PR present X Layer USDT0 as a working option — and right now it isn't.

**Setup:** funded a local wallet with $0.067 USDT0 on X Layer (`0x2d4B…E2F6`), 0 USDC on Base, ran an unauthenticated `smart-money/netflow` call so the CLI iterates the `accepts` list (Base → X Layer → Solana).

**Result:** every X Layer signature is rejected by the API facilitator with `error: invalid_exact_evm_signature`, and the rejection echoes our payload's `extra.name` back as `"USDâ®0"` — which is exactly what you get when UTF-8 `₮` (`0xE2 0x82 0xAE`) is decoded as Latin-1 (`â‚®`). So the facilitator is treating `extra.name` as Latin-1 when re-deriving the EIP-712 domain hash, and rejecting signatures the on-chain contract would otherwise accept.

**Confirmed the CLI side is correct:**

- The EIP-712 `DOMAIN_SEPARATOR` derived from `name="USD₮0"`, `version="1"`, `chainId=196`, `verifyingContract=0x779Ded0c9e…713736` matches the on-chain `DOMAIN_SEPARATOR()` byte-for-byte (`0xd591d9baf744328d9400b923cb02c9474d367d591ca1ab24d8c4068be527599d`).
- Token supports EIP-3009 (`transferWithAuthorization` selector reverts with `TetherToken: auth expired` for zero-arg call → function exists).
- Base USDC (`name="USD Coin"`, ASCII) signs and is accepted on the same flow with no changes.

**Action:** server-side facilitator needs to read `extra.name` as UTF-8 when computing the EIP-712 domain hash. Until that's fixed, the X Layer rail in `accepts` is unsignable in practice.

I can file this as a Linear issue against the API team if that's the right venue — let me know.

gulshngill:
Addressed comment 4333223210: corrected the X Layer USDT0 token address in src/x402.js to 0x779Ded0c9e1022225f8E0630b35a9b54bE713736 and updated the inline comment.

gulshngill:
Addressed comment 4333224751: documented X Layer USDT0 caveats while the upstream issue was being investigated.

**Update (2026-04-29):** upstream issue resolved; warnings removed in 99b22c9. X Layer USDT0 is now a regular supported x402 option alongside Base USDC and Solana SPL USDC.

gulshngill:
## Independent Review — follow-up

The address fix from comment `4333223210` (X Layer USDT0 → `0x779Ded0c9e1022225f8E0630b35a9b54bE713736`) is verified in the tree — no instances of the old wrong address remain.

The warnings from comment `4333224751` have since been **removed** in 99b22c9 — the upstream issue they were tracking has been resolved, so X Layer USDT0 is now a regular supported x402 option.

**Follow-ups status:**
1. ✅ `AGENTS.md` cleaned up — X Layer USDT0 is now listed alongside Base/Solana with no caveats.
2. ✅ `src/api.js` 402 error message cleaned up — lists USDT0 on X Layer as a valid funding option.
3. ✅ Low-balance warning now uses dynamic symbol (`${symbol} balance low ($X.XX)`) — X Layer wallets are no longer told their "USDC" is low.
4. ✅ Unit-test coverage added for `EVM_NETWORKS` in `checkX402Balance()` — `src/__tests__/x402.test.js` (53ab205) pins the Base USDC + X Layer USDT0 contracts and the unknown-network fallback. The X Layer assertion uses a literal contract address so a typo regresses the test instead of silently the prod path.

Lint and `npm test` both clean (1398 passing, 2 skipped).

gulshngill:
Pushed [`93e6a6d`](https://github.com/nansen-ai/nansen-cli/commit/93e6a6d):

**Dynamic stablecoin symbol in low-balance warning** (`src/x402.js`, `src/api.js`)
- `checkX402Balance(network)` now returns `{ balance, symbol }` instead of just a number, with `symbol` set per-network (USDC for Base/Solana, USDT0 for X Layer).
- The warning in `_x402Retry` now reads `${result.symbol} balance low ($X.XX)` so X Layer wallets are no longer told their "USDC" is low.

_(This commit also added temporary X Layer warnings while we were investigating an upstream issue; those have since been reverted in 99b22c9 now that the issue is resolved. The dynamic-symbol fix above survived.)_

Lint and `npm test` both clean (1394 passing, 2 skipped).

0xlaveen:
**Correction to my earlier comment ([#issuecomment-4333224751](https://github.com/nansen-ai/nansen-cli/pull/422#issuecomment-4333224751)) — root cause is client-side, not server-side**

Apologies for the wrong attribution. After the server-side commits landed I retested and the X Layer rail was still rejecting with `invalid_exact_evm_signature`. Tracing it, the bug is in this PR's CLI code, not the API:

**`src/x402.js:27`** uses `atob()` to decode the base64 `Payment-Required` header. `atob()` returns a Latin-1 binary string — every UTF-8 byte becomes its own UTF-16 code unit. So when the server sends `extra.name = "USD₮0"` (`₮` is `0xE2 0x82 0xAE` in UTF-8), the CLI parses it as `"USDâ®0"` (`U+E2 U+82 U+AE`), then signs the EIP-712 payload against a domain whose `name` doesn't match the on-chain token. Server-side verifier (correctly) rejects.

Base USDC works because `"USD Coin"` is pure ASCII — `atob` is a no-op for ASCII.

**Proof (DOMAIN_SEPARATOR comparison):**
```
on-chain      DOMAIN_SEPARATOR: 0xd591d9baf744328d9400b923cb02c9474d367d591ca1ab24d8c4068be527599d
atob (today)  DOMAIN_SEPARATOR: 0xccd22af6a62860a191b3edd7560b47790df8a686b7b6757e524b9e030c176ac4  ← MISMATCH
utf-8 (fixed) DOMAIN_SEPARATOR: 0xd591d9baf744328d9400b923cb02c9474d367d591ca1ab24d8c4068be527599d  ← MATCH
```

**One-line patch:**
```diff
-    const decoded = JSON.parse(atob(header));
+    const decoded = JSON.parse(Buffer.from(header, 'base64').toString('utf8'));
```

**End-to-end verified after applying locally:**

- POST `/api/v1/smart-money/netflow` with no API key → CLI tries Base USDC (rejected, wallet empty) → tries X Layer USDT0 → **accepted, status 200**
- Wallet balance: `$0.067 USDT0 → $0.017 USDT0` (delta = `amount: 50000` at 6 decimals = $0.05 exactly)
- On-chain tx: [`0xe4f45764…d5777`](https://www.okx.com/web3/explorer/xlayer/tx/0xe4f45764f815789e5c8212eab6b3b3846dda98f0173d24e39978e94b4e8d5777) settled in block 58689840, facilitator paid gas, user signed only the EIP-3009 auth
- Real Smart Money netflow data returned (BURNIE, PUMP, GRASS…)

**Implication for the recent commits:** `eae1939` and `99b22c9` removed the X Layer "temporarily unavailable" caveats on the assumption that the server side had been fixed. The rail is in fact still broken on `main` of this PR — the docs only become accurate once the `atob` → UTF-8 fix above ships. Worth either bundling that fix here, or keeping a temporary caveat in the docs until it does.

Sorry again for the bad initial attribution.

gulshngill:
@0xlaveen — you nailed the root cause. `atob()` returns a Latin-1 binary string, so the `₮` byte sequence (0xE2 0x82 0xAE) in `extra.name = 'USD₮0'` was getting parsed as `'USDâ®0'`, then signed against the wrong EIP-712 domain, then correctly rejected by the server. Base USDC worked because `'USD Coin'` is pure ASCII; X Layer USDT0 was the only payload that hit the bug.

Fixed in 8f9397f by switching to `Buffer.from(header, 'base64').toString('utf8')` (also fixed a second `atob()` site in the WalletConnect path in `src/api.js`, plus a regression test for UTF-8 decode).

Worth flagging: the earlier commits eae1939 and 99b22c9 that removed the X Layer "temporarily unavailable" warnings were premature — those docs are only accurate once **this** fix ships. So the docs and the code should land together as one merge, not separately.

gulshngill:
Fresh review of PR #422 after the latest UTF-8/x402 fixes.

I reviewed the PR against `origin/main...HEAD` (local `main` is stale in this checkout, so the requested `git diff main...docs/mpp-tempo-guidance -- src/` is noisy and includes unrelated older changes). I also ran the requested `grep -rn 'atob' src/` and the full test suite.

## Finding fixed during review

I found one remaining UTF-8 issue in the WalletConnect x402 path:

- `src/api.js` now decodes `Payment-Required` with `Buffer.from(paymentHeader, 'base64').toString('utf8')`, which is correct.
- But `src/walletconnect-x402.js` still re-encoded the final `Payment-Signature` payload with `btoa(JSON.stringify(paymentPayload))`.
- Once `accepted.extra.name` is correctly decoded as `USD₮0`, that `btoa()` call throws `InvalidCharacterError` in Node because `₮` is outside Latin-1. So WalletConnect/Privy x402 signing could still fail for X Layer USDT0 even though the header decode was fixed.

I pushed `aaf109b` to fix this:

- Replaced WalletConnect payment payload encoding with `Buffer.from(JSON.stringify(paymentPayload), 'utf8').toString('base64')`.
- Added API-path regression coverage asserting the WalletConnect fallback receives `USD₮0` intact from the decoded `Payment-Required` header.
- Added payload-encoding regression coverage asserting `buildPaymentSignatureHeader()` preserves `accepted.extra.name = 'USD₮0'`.
- Updated the UTF-8 changeset to mention both decode and WalletConnect payload encode.
- Tightened Tempo docs/error guidance from plain `tempo wallet` to `tempo wallet login` / `tempo wallet whoami`.

## Review checks

- `grep -rn 'atob' src/` now only finds test-only decode helpers and comments. There are no runtime `atob(` calls left. I also checked runtime `btoa(`; none remain under `src/` outside tests after `aaf109b`.
- The second `src/api.js` decode fix is correct: it now UTF-8 decodes before passing requirements into `handleX402Payment()`, and the new test covers the non-ASCII X Layer metadata path.
- UTF-8 regression coverage is now sufficient: `parsePaymentRequirements()`, the `src/api.js` WalletConnect fallback decode path, and the WalletConnect/Privy `Payment-Signature` payload encoder are all covered.
- X Layer USDT0 address is consistent everywhere it matters: `0x779Ded0c9e1022225f8E0630b35a9b54bE713736`. The old bad address only remains as a regression-test comment.
- `checkX402Balance()` refactor is safe: the only runtime caller is `_x402Retry()`, which now reads `{ balance, symbol }`; tests cover Base USDC, X Layer USDT0, unknown EVM fallback, and RPC failure.
- Changesets look correct: minor for X Layer support, patches for balance-warning, UTF-8, and MPP docs. The UTF-8 changeset was updated for the WalletConnect encode fix.
- README / AGENTS / `skills/nansen-mpp-payment` now accurately describe three auth rails, keep MPP out of `nansen-cli` proper, and no longer carry the stale X Layer caveat.

## Verification

- `npm test` passed: 25 files, 1401 passed, 2 skipped.
- `npm run lint` passed.

Verdict: APPROVE. The only issue I found was fixed and pushed in `aaf109b`; I do not see remaining blockers for merge.

PR #441, opened 2026-05-18
fix(research): warn when unsupported --sort or --page/--limit flags are passed

kome12:
## Summary

Two \`nansen research historical-*\` subcommands silently ignored flags without any feedback to the caller:

- **\`historical-token-flow-summary\`** accepted \`--page\`/\`--limit\` in its CLI options but the handler never forwarded them to the API. Callers had no indication their pagination flags were being dropped.
- **\`historical-smart-money-balances\`** accepted \`--sort\` in its CLI options but the handler never forwarded it to the API. Callers had no indication their sort flag was being dropped.

## Fix

Added \`process.stderr.write()\` warnings (matching the existing \`⚠️\` pattern used elsewhere in the codebase) that fire when the unsupported flag is detected:

- \`historical-token-flow-summary\` + \`--page\`/\`--limit\` → warns: \`⚠️  warning: --page/--limit is ignored by historical-token-flow-summary (endpoint does not support pagination)\`
- \`historical-smart-money-balances\` + \`--sort\` → warns: \`⚠️  warning: --sort is ignored by historical-smart-money-balances (endpoint does not support order_by)\`

Warnings go to stderr so they don't pollute JSON stdout output. The API call still succeeds — callers get data, plus a clear signal that the flag had no effect.

## Tests

Added two regression tests in \`src/__tests__/research.test.js\`:

1. Calls \`historical-token-flow-summary\` with \`--page 2 --limit 5\` and asserts a matching warning was written to stderr.
2. Calls \`historical-smart-money-balances\` with \`--sort value_usd:desc\` and asserts a matching warning was written to stderr.

Both new tests pass alongside all 1475 existing tests (2 pre-existing skips unchanged).

## Affected flag combinations

| Subcommand | Silently ignored flag | Now warns |
|---|---|---|
| \`historical-token-flow-summary\` | \`--page\`, \`--limit\` | ✓ |
| \`historical-smart-money-balances\` | \`--sort\`, \`--order-by\` | ✓ |

### DIFF
--- a/.changeset/research-warn-unsupported-flags.md
+++ b/.changeset/research-warn-unsupported-flags.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
+Warn to stderr when --page/--limit is passed to historical-token-flow-summary or --sort is passed to historical-smart-money-balances, since those endpoints silently ignore those flags.
--- a/src/__tests__/research.test.js
+++ b/src/__tests__/research.test.js
@@ -315,6 +315,20 @@ describe('buildResearchCommands handler', () => {
     expect(mockApi.researchTokenFlowSummary).toHaveBeenCalled();
   });
 
+  it('warns to stderr when --page/--limit is passed to historical-token-flow-summary', async () => {
+    mockApi = makeMockApi();
+    const stderrSpy = vi.spyOn(process.stderr, 'write').mockImplementation(() => {});
+    try {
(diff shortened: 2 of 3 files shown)


kome12:
Closing — the fix approach was wrong. The correct fix is to reject the unsupported flags with a clear error rather than silently warning, since these flags have no effect at the API level.

PR #457, opened 2026-06-11
feat(x402): BSC multi-asset payments + Permit2 signing

gulshngill:
## What

The Nansen API now advertises **four stablecoins on BNB Smart Chain** — U (United Stables), USD1, USDT, and USDC. This PR makes the CLI able to pay with all of them.

U and USD1 already work with the CLI's existing EIP-3009 flow (the server lists them first for exactly that reason). USDT and USDC on BSC predate EIP-3009, so facilitators settle them through Uniswap's canonical [Permit2](https://github.com/Uniswap/permit2) contract — which needs a different signature type.

## Changes

**`src/x402-evm.js` — Permit2 signing (`permit2-exact`)**
- Payment creation routes on the 402's `extra.assetTransferMethod`: absent/`eip3009` → existing gasless flow; `permit2-exact` → new `createPermit2ExactPayload`; anything else throws so the fallback loop moves to the next accepts entry.
- Signs a `PermitWitnessTransferFrom` against the spender contract advertised in the 402 (`extra.spenderAddress` — never hardcoded), witness bound to the merchant wallet. Nested-struct EIP-712 hashing and the 3-field Permit2 domain are implemented in the repo's existing zero-dependency style.
- **Cross-validated**: the digest test vector was computed with an independent EIP-712 implementation (Python `eth_account`) over identical input — the two agree byte-for-byte.

**`src/x402.js` — allowance preflight + multi-asset registry**
- Permit2 requires a one-time on-chain `token.approve(PERMIT2_ADDRESS, …)` from the payer. Entries are skipped with an actionable message when the allowance is missing, instead of burning a doomed verify round-trip. RPC failures fail open (server decides).
- `EVM_X402_TOKENS` becomes an ordered per-network token list (BSC: U, USD1, USDT, USDC — all 18-decimal BEP-20 deployments); RPC URLs split into `EVM_X402_RPCS`.
- `checkX402Balance(network, asset)` checks the token actually paid with, so low-balance warnings use the right symbol and decimals.

**`src/api.js`** — threads the paid asset through to the balance warning.

## Testing

- New `x402-permit2.test.js` (7 tests): reference digest vector, payload wire shape (decimal-string numerics, no EIP-3009 fields), missing-spender error, method routing (permit2-exact / eip3009 / absent / unsupported).
- Registry tests updated for the multi-asset shape.
- Full suite: 1498 passed.

## Notes

- Wallets holding U or USD1 pay on BSC with no extra steps. Wallets paying USDT/USDC on BSC need the one-time Permit2 approval (requires a little BNB for that single transaction; every payment after is a gasless signature).
- `permit2-upto` is intentionally unsupported for now — the CLI only uses exact-amount payments.

### DIFF
--- a/src/x402.js
+++ b/src/x402.js
@@ -59,13 +59,14 @@ function rankRequirements(requirements) {
 const ALLOWANCE_SELECTOR = '0xdd62ed3e';
 
 /**
- * Check whether `owner` has approved Permit2 to spend `token`.
- * Permit2-based payments are doomed without this one-time on-chain approval,
- * so skip those entries early instead of burning a failed verify round-trip.
+ * Check whether `owner` has approved Permit2 to spend at least `amount` of
+ * `token`. Permit2-based payments are doomed without sufficient allowance
+ * (never approved, or a finite approval now below the payment amount), so
+ * skip those entries early instead of burning a failed verify round-trip.
  * Returns true when the allowance is unknown (RPC failure) — let the server
  * decide rather than block a possibly-valid payment.
  */
-async function hasPermit2Allowance(network, token, owner) {
+async function hasPermit2Allowance(network, token, owner, amount) {
   const rpc = getEvmRpcUrl(network);
   if (!rpc) return true;
   try {
@@ -82,7 +83,7 @@ async function hasPermit2Allowance(network, token, owner) {
     });
     const data = await resp.json();
     if (typeof data.result !== 'string') return true;
-    return BigInt(data.result) > 0n;
+    return BigInt(data.result) >= BigInt(amount);
   } catch {
     return true;
   }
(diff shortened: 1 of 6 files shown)


src/x402.js
```diff
@@ -55,12 +55,60 @@ function rankRequirements(requirements) {
...
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        jsonrpc: '2.0', id: 1,
+        method: 'eth_call',
+        params: [{ to: token, data: `${ALLOWANCE_SELECTOR}${ownerArg}${spenderArg}` }, 'latest'],
+      }),
+    });
+    const data = await resp.json();
+    if (typeof data.result !== 'string') return true;
+    return BigInt(data.result) > 0n;
```
kome12: Minor: this preflight only detects the **zero-allowance** case (`> 0n`), but the skip message below tells users to `approve(Permit2, <amount>)`. A wallet that approved a small finite allowance now exhausted (or below the payment amount) would pass this check and then fail server-side instead of getting the actionable skip message.

Since it fails open on RPC error anyway, the heuristic is defensible — but consider either a one-line comment noting it only catches the never-approved case, or threading the payment amount through to compare `allowance >= amount`. Non-blocking.

kome12 (dismissed):

src/x402.js
gulshngill: Good catch — went with the real fix rather than the comment: `hasPermit2Allowance` now takes the payment amount and compares `allowance >= amount`, so the exhausted/insufficient-finite-approval case gets the actionable skip message too. Skip message updated to name both cases. Fail-open on RPC error unchanged. Fixed in d02572c.

PR #469, opened 2026-07-29
feat: surface the API's credit and rate-limit response headers

gulshngill:
The API reports what a call cost and what quota is left on every response. The client dropped those headers on the floor, so the only credit figure the CLI could show was the static per-endpoint estimate published in the API reference (`src/cost-cache.js`, rendered in `--help` as `Cost: N credits (Free tier) / M credits (Pro tier)`) — a quote, never what you were actually charged.

## What changes

New `src/response-meta.js` parses the headers defensively. Every field is optional: some auth rails charge no credits, some responses are served before quota is resolved, and older deployments may not send the rate-limit triplet at all — so a **missing header means unknown, never zero**.

**Failures carry the numbers.** `NansenError.details` gains `credits` and `rateLimit`, and `formatError()` already surfaces `details`, so this needed no new plumbing. An out-of-credits error now reports the real remaining balance; a rate-limited one reports the window state.

```json
{ "success": false, "error": "Insufficient credits", "code": "CREDITS_EXHAUSTED", "status": 403,
  "details": { "credits": { "used": 0, "remaining": 2 } } }
```

**Low balance warns on stderr.** Fired only when the balance will not cover another call of the size just made (`remaining < used`, or `remaining === 0`). That is self-scaling across plans — no absolute threshold to tune, and it stays silent until it is actually actionable, which matters because a free-tier user at 10 credits/day and a Pro user share the code path. stderr keeps stdout pure JSON for agents, matching the existing `⚠️` warnings in `src/cli.js`.

**Successes expose the same data without changing output.** Two accessors, because they fail in different ways:

- `result[RESPONSE_META]` — a **Symbol**, so `JSON.stringify` and `Object.keys` both skip it. Race-free and exact, but lost if a command handler reshapes the body (`filterFields`, the local filtering in `token screener --search`).
- `api.lastResponseMeta` — survives any reshaping. Last write wins if a handler makes several calls, which for a balance warning is the right answer.

## Why the output shape is untouched

`formatOutput()` JSON-stringifies the whole object, so attaching a plain `data._meta` would have changed what **every** command prints. The symbol avoids that entirely — a test asserts `JSON.stringify(result)` is byte-identical and `Object.keys(result)` is unchanged.

## Compatibility

Most existing success-path mocks in `src/__tests__/api.test.js` are `{ ok: true, json: async () => ... }` with **no `headers` at all**. `readResponseMeta` uses `response?.headers?.get?.(name)` and returns `null` rather than throwing, and the header bag only needs a `.get()` — a real `Headers` or a `Map`, as the existing tests use. Attachment also happens *after* the response cache is written, since quota numbers are per-response and would be stale on a cache hit.

## Testing

- `npm test` — 1528 passed, 2 skipped, 0 failed
- `npm run lint` — clean
- 16 new tests in `src/__tests__/response-meta.test.js`: parsing, zero-vs-absent, unparseable values, the warning thresholds, the JSON-invisibility guarantee, and the 403/429 error paths

`src/schema.json` needs no update — no commands or options were added.

## Deliberately not here

The API is about to grow a per-call cost header, a request id, and a stable machine-readable error `code`. That last one would let us delete the prose string-matching in `statusToErrorCode()` (`messageLower.includes('credit')`, `.includes('address')`, `.includes('chain')`…), which is fragile — those messages can be reworded server-side at any time and CLI error codes would silently drift. Left for a follow-up once that surface ships, rather than guessing at it now.

### DIFF
--- a/.changeset/surface-credit-and-rate-limit-headers.md
+++ b/.changeset/surface-credit-and-rate-limit-headers.md
@@ -1,0 +1,11 @@
+---
+"nansen-cli": minor
+---
+
+Surface the API's credit and rate-limit response headers.
+
+Failed calls now report quota state in their error details: an out-of-credits error carries your actual remaining balance, and a rate-limited error carries the limit, what is left, and how long the window needs to drain. Previously the only credit figure the CLI could show was the static per-endpoint estimate published in the API reference — a quote, not what you were charged.
+
+A warning goes to stderr when your balance will not cover another call of the size just made, so it never interferes with the JSON on stdout.
+
+Successful responses carry the same numbers under an exported `RESPONSE_META` symbol, and the client exposes `lastResponseMeta`. Both are additive: the JSON each command prints is unchanged.
--- a/README.md
+++ b/README.md
@@ -177,17 +177,19 @@ nansen research smart-money netflow --chain solana --fields token_symbol,net_flo
 
 ```json
 { "success": true,  "data": <api_response> }
-{ "success": false, "error": "message", "code": "ERROR_CODE", "status": 401 }
+{ "success": false, "error": "message", "code": "ERROR_CODE", "status": 401, "details": { ... } }
 ```
 
 **Critical error codes:**
 
 | Code | Action |
 |------|--------|
-| `CREDITS_EXHAUSTED` | Stop all API calls immediately. Check [app.nansen.ai](https://app.nansen.ai). |
+| `CREDITS_EXHAUSTED` | Stop all API calls immediately. `details.credits.remaining` is your actual balance. Top up at [app.nansen.ai/api](https://app.nansen.ai/api). |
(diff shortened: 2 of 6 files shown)


Codier (dismissed):
Reviewed the full diff. No correctness defects found.

Checked and cleared: header casing (lowercase constants against case-insensitive `Headers.get`), missing/undefined header handling, primitive and array bodies, the output-shape invariant (symbol is skipped by `JSON.stringify` and `Object.keys`), and error-path precedence (header-derived values are spread after `...data`, so a body field can't shadow them). Attaching after the cache write is correct — quota numbers are per-response.

CI verified job by job: lint, test on Node 20/22/24, and Aikido all pass; publish and sync-clawhub are skipped by design. The BLOCKED merge state was the missing required review, not CI.

Approving.

PR #470, opened 2026-07-29
feat: surface the API's request id

gulshngill:
**Stacked on #469** — base is `feat/surface-credit-and-rate-limit-headers`, not `main`, so the diff here is just the incremental change. `src/response-meta.js` only exists on that branch. Merge #469 first; GitHub will retarget this to `main` automatically.

## Why

A server error was effectively unreportable. Nothing identifying the failed request ever reached the user, so a bug report could only describe symptoms — no way to find the call in our logs. The API now returns an id on every response; this reads it.

## What changes

`details.requestId` on failures. `formatError()` already surfaces `details`, so no new plumbing:

```json
{ "success": false, "error": "Internal server error", "code": "SERVER_ERROR",
  "status": 500, "details": { "requestId": "req-a1b2c3..." } }
```

Successes carry it alongside the credit and rate-limit figures under the existing `RESPONSE_META` symbol, so default JSON output is still unchanged.

Also factored the header reader into `stringHeader` / `intHeader`, since the id is the first non-numeric header this parses.

## Three decisions worth reviewing

**Opaque, never parsed.** Trimmed, and that is all. Gateway-issued, caller-forwarded and generated ids all look different, and the format is not ours to depend on. A test pins that any format survives untouched.

**Omitted, not `null`, when the header is absent.** This lets the CLI ship ahead of the API deploy — on a deployment that does not send the header, behaviour is exactly what it is today. Consistent with how `credits` and `rateLimit` already behave.

**Last attempt's id on a retried call.** Each attempt gets its own id server-side; the last one is the failure actually being reported. Noted in a comment at the assignment site so it does not read as an oversight.

## Testing

- `npm test` — 1532 passed, 2 skipped, 0 failed
- `npm run lint` — clean
- 4 new tests: parsing, the opaque/whitespace contract, the 500 path, and the absent-header case

While adding the last one I rewrote a `.catch(assert)` into an explicit capture — the original form would have passed silently had the call resolved instead of rejecting.

## Not here

The CLI could **generate and forward** its own `X-Request-Id` — the API echoes a forwarded id back unchanged. That is strictly better for support, because on a timeout or dropped connection there is no response at all, so a server-issued id is unquotable exactly when you most need one. Deliberately a separate change; it alters what we send rather than what we read.

### DIFF
--- a/.changeset/surface-request-id.md
+++ b/.changeset/surface-request-id.md
@@ -1,0 +1,9 @@
+---
+"nansen-cli": minor
+---
+
+Surface the API's request id.
+
+Failed calls now carry `details.requestId` — the value that identifies the call end to end. Quote it when reporting a problem; previously nothing identifying a failed request ever reached the user, which made server errors effectively unreportable. Successful responses expose it alongside the credit and rate-limit figures under the `RESPONSE_META` symbol.
+
+Absent on deployments that do not send the header yet, in which case the field is simply omitted.
--- a/README.md
+++ b/README.md
@@ -188,8 +188,17 @@ nansen research smart-money netflow --chain solana --fields token_symbol,net_flo
 | `UNAUTHORIZED` | Wrong or missing key. Re-auth. |
 | `RATE_LIMITED` | Auto-retried by CLI. `details.rateLimit.resetSeconds` is how long the window needs to drain. |
 | `UNSUPPORTED_FILTER` | Remove the filter and retry. |
+| `SERVER_ERROR` | Not your fault. Quote `details.requestId` when reporting it. |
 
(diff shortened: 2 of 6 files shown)


Codier (dismissed):
Reviewed the full incremental diff. No correctness defects found.

Verified: `stringHeader` refactor is behavior-preserving for the existing int headers; absent header omits `requestId` rather than setting null, so shipping ahead of the API deploy is a true no-op; `formatError()` carries `details.requestId` through to the user as described; the retried-call path evaluates the id fresh per attempt, so a headerless final attempt can't inherit a stale one. The id is an opaque server-issued correlation token — no credential or user data.

One thing worth flagging on the process side rather than the code: GitHub Actions never ran on this branch. `.github/workflows/ci.yml` filters `pull_request.branches: [main]`, and this PR is based on the #469 stack branch, so lint/test never triggered — only Aikido (an external App, not branch-filtered). I validated the diff locally instead: lint clean, 1532 passed / 2 skipped / 0 failed on Node 22 — but that's one leg, not the 20/24 matrix. Worth confirming a CI run appears after #469 merges and GitHub retargets this to main.

Approving.

PR #485, opened 2026-08-11
BI-981: surface Hyperliquid order id in perp order/close output

MarcLlopart:
Parses the order id (`oid`) and fill from Hyperliquid's `/exchange` response and surfaces them after `nansen perp order` and `perp close` — the perp analogue of how spot trading prints its `quoteId` — and emits them to BI as a new `perp_order_completed` telemetry event.

## Why

After placing a perp order the CLI dumped the raw HL response JSON, so the `oid` a user needs to track or `cancel` an order was buried. This makes it first-class, and copy-paste actionable. The same parsed outcome is also sent to BI (`admin_core.tracks`) so we can analyse fill rates, slippage, and TP/SL adoption — data that lived only in the user's terminal before.

## What

- New exported `summarizeOrderResult(result, action)` in `src/perp.js` — parses `response.data.statuses[]` into `{leg, kind, oid, totalSz, avgPx}`. Gated on the *submitted* action being an `order`, so leverage / transfer / builder-fee / cancel actions (type `default`) keep the concise output.
- `buildScreenSignSubmit` now prints:
  - **Filled:** `Filled: 0.01 @ 2000.5  (oid 55)`
  - **Resting:** `Resting order: oid 987654` + `Cancel:  nansen perp cancel --coin ETH --oid 987654`
- TP/SL bracket legs are labelled `parent` / `take-profit` / `stop-loss` (same logic as `extractActionErrors`), each printing its own `oid`.
- `coin` threaded through the `order`/`close` handlers so the cancel hint is copy-pasteable.

## Design note

Client-side only — reads the `oid` straight from the `/exchange` response. **No Nansen/superapp route is involved**, consistent with the shipped "order path stays direct-to-HL" design (Decision D4).

## Order-outcome telemetry (`perp_order_completed`)

Builds on `summarizeOrderResult` above: `buildScreenSignSubmit` now also emits an anonymous `perp_order_completed` event so these outcomes reach BI. `cli_command_succeeded` fires at the command wrapper *before* the HL response, so it never sees the fill; the order path bypasses the Nansen API on submit (D4), so the backend never sees it either — this client-side event is the only place order outcomes are observable.

- Fired from `buildScreenSignSubmit` right after `summarizeOrderResult`, reusing its parsed legs. **order/close only** — cancel / leverage / transfer / builder-fee carry no telemetry descriptor and also summarize to `[]`.
- Payload: `command`, `coin`, `side` (buy/sell), `order_type` (market/limit), `oid`, `status` (filled/resting), `fill_price`, `fill_size`, `requested_price`, `requested_size`, `has_tp_sl`, and `tp_sl_legs[]` (`{leg, status, oid, fill_price?, fill_size?}`) for brackets. Partial fills surface as `fill_size < requested_size`.
- Wrapped in `try/catch` and swallowed so a telemetry failure can never downgrade a real fill into `cli_command_failed`; `await`ed so the event flushes before the process exits.
- Honours the existing `DO_NOT_TRACK` / `NANSEN_NO_TELEMETRY` opt-out. Order rejections stay on `cli_command_failed` (`submitExchange` throws before the event).
- `trackPerpOrderCompleted` is injectable via `buildPerpCommands` deps (defaults to the real tracker) for testability.

_Staging check still pending a human: place a testnet / low-value order and confirm `perp_order_completed` lands in `admin_core.tracks` with `oid` matching the terminal output._

## Tests

- Output: +7 tests (`src/__tests__/perp.test.js`) — 5 unit tests for `summarizeOrderResult` (resting, filled, TP/SL bracket labelling, non-order → `[]`, no-statuses → `[]`) and 2 integration tests asserting the `order` handler prints the oid + cancel hint (resting) and the fill with no cancel hint (market).
- Telemetry: +5 (`src/__tests__/telemetry.test.js`: event shape, `/perp/close` path, resting nulls, TP/SL legs, opt-out) and +7 (`src/__tests__/perp.test.js`: market fill, resting limit, partial fill, bracket legs, close command, non-order actions don't fire, rejection doesn't fire).

## Verification

- `npm test` → 1885 passed / 2 skipped (51 files)
- `npm run lint` → clean
- No `schema.json` change (output-only; no new commands/options)
- Changesets: `minor` (order-id output) + `patch` (`perp_order_completed` telemetry); trading skill doc updated


gulshngill:
## Independent Codex review

Actionable findings, ranked by severity:

1. **High — valid `uint64` order IDs can be rounded before the new cancel command is printed** (`src/hl-client.js:115-118`, `src/perp.js:272-276`, `src/perp.js:346-349`, `src/perp.js:468-480`). Hyperliquid defines `oid` as a `uint64` ([official API docs](https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/info-endpoint#query-order-status-by-oid-or-cloid)), but `JSON.parse()` and `parseInt()` convert it to a JavaScript `Number`. Any valid OID above `Number.MAX_SAFE_INTEGER` is silently changed (for example, `9007199254740993` becomes `9007199254740992`), so this PR can print and submit a cancel for the wrong order while presenting it as copy-paste actionable. Preserve OIDs losslessly through response decoding and cancel-action construction (string/`BigInt` with compatible wire serialization), or at minimum detect unsafe values and withhold the cancel hint; add a round-trip test above `2^53`.

2. **Medium — the existing telemetry disclosure no longer describes what the CLI sends** (`src/telemetry.js:4-6`, `src/telemetry.js:303-316`, `src/cli.js:774`). Users are currently told only that “anonymous usage stats” are collected, and the module says those stats cover commands, duration, and errors. This PR adds exact trade-level fields—OID, symbol, side, requested/fill price and size, and TP/SL outcomes—linked to the persistent `anonymous_id`. That is a material expansion beyond the disclosed scope even if the transport remains pseudonymous and opt-out. Update the user-visible disclosure/docs to name the trade details before shipping (or minimize/gate these fields separately), and test the revised disclosure/opt-out contract.

CI is green and the new unit tests cover the documented small-OID response shapes, output, telemetry payloads, failures, and `DO_NOT_TRACK`; they do not cover lossless OID handling or the expanded disclosure.

MarcLlopart:
Addressed the independent review findings in `d3d36ba`.

## High — uint64 order id precision
Hyperliquid `oid` is a uint64, but `JSON.parse()` in `hl-client.js` already narrows it to a JS `Number`, so any id above 2^53 arrives rounded (the value is lost before the parser returns — a reviver cannot recover it). Rather than re-architect JSON/msgpack across the read/cancel money-path for ids that are ~10 digits today, this takes the reviewer's "at minimum" path:

- `summarizeOrderResult` now flags each leg with `oidSafe = Number.isSafeInteger(oid)`.
- Output **withholds the exact id and the copy-paste `perp cancel`** for an unsafe id (prints `oid too large to display precisely`), so a wrong id is never presented as actionable.
- Telemetry **omits** the parent oid / **nulls** leg oids when unsafe, so BI records "unknown" instead of a rounded id.
- Added round-trip tests at the 2^53 boundary (unit + output + telemetry).

Full lossless handling (string/BigInt through parse → cancel wire → msgpack) remains a larger, separate change worth doing if HL oids ever approach 2^53 — happy to file a follow-up issue.

## Medium — telemetry disclosure
The disclosure now names the trade-level fields `perp_order_completed` sends instead of describing only commands/timing/errors:

- CLI help footer (`cli.js`): now lists asset, side, order type, sizes, prices, order id, and TP/SL outcomes, tied to a random anonymous id.
- `telemetry.js` module doc: same, with the opt-out env vars.
- Added an opt-out contract test asserting `DO_NOT_TRACK` / `NANSEN_NO_TELEMETRY` suppresses the event.

## Verification
`npm test` → 1889 passed / 2 skipped · `npm run lint` clean.

MarcLlopart:
Follow-up on the Medium (telemetry) finding — `21b109b`.

Rather than disclose the full trade-level payload, the `perp_order_completed` event is now **minimized** to just the **trade side + Hyperliquid order id** (plus the standard envelope: anonymous id, session, timestamp, path, CLI version). Dropped: asset, order type, requested/filled price & size, fill status, and TP/SL leg outcomes.

A trade (fill) id isn't carried by the order-placement response — it only exists once the order fills, via the fills feed — so side + oid is the reliable maximum. The oid is still omitted when it exceeded JS safe-integer precision. The disclosure (help footer + `telemetry.js` doc) now names exactly these two fields, and the console output of `perp order`/`close` is unchanged (still shows fills/oid/legs) — only the BI event shrank.

`npm test` → 1886 passed / 2 skipped · `npm run lint` clean.

gulshngill (approved):

PR #493, opened 2026-08-15
fix: validate --slippage-bps on limit-order create

crazywriter1:
## Summary
- Validate `--slippage-bps` on `limit-order create` with the same 0–10000 bounds already used by `update` (and bridge)
- Reject invalid values client-side before auth/deposit/API calls
- Align schema/help text with the enforced range
- Add regression tests for out-of-range and valid create input

## Test plan
- [x] `npx vitest run src/__tests__/limit-order.test.js -t slippage`
- [x] `nansen trade limit-order create ... --slippage-bps 15000` → error, exits 1
- [x] `nansen trade limit-order create ... --slippage-bps 100` → accepted (with other required flags)
- [x] Omit `--slippage-bps` → still allowed (auto)

### DIFF
--- a/src/limit-order.js
+++ b/src/limit-order.js
@@ -401,6 +401,17 @@ export function parseExpiry(expiryStr) {
   throw new Error(`Invalid expiry format: "${expiryStr}". Use "24h", "7d", "30d", or epoch ms.`);
 }
 
+// Whole integer bps in [0, 10000], matching bridge parseSlippageBps.
+// Number() would accept "1.5", "1e2", "0x10", and boolean true.
+function parseSlippageBps(raw) {
+  const s = String(raw).trim();
+  const bad = 'Error: --slippage-bps must be a whole integer between 0 and 10000 basis points.';
+  if (!/^\d+$/.test(s)) throw new Error(bad);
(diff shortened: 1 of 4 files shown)


crazywriter1:

src/limit-order.js
```diff
@@ -581,6 +581,18 @@ EXAMPLES:
         return;
       }
 
+      // Same bounds as update / bridge: whole-order slippage is 0–10000 bps.
+      let slippageBps;
+      if (slippageBpsRaw != null) {
+        const bps = Number(slippageBpsRaw);
```
kome12: Do we want to make this stricter so it enforces integer digits? Right now I can pass in `1.5`, `1e2`, `0x10` and `true` and it won't give me the `Error: --slippage-bps must be between 0 and 10000 basis points.`

crazywriter1: Good catch. Tightened create and update to digits-only whole integers in 0–10000 (same shape as bridge parseSlippageBps). 1.5 / 1e2 / 0x10 / true now fail with a clear error.

kome12 (approved):

PR #495, opened 2026-08-19
Verify EVM swap outcome via balance-delta simulation

kome12:
## What

Adds outcome verification to `trade execute` as defence-in-depth on top of the existing pre-broadcast checks. Before broadcasting an EVM swap, the CLI simulates the transaction through a trace-capable endpoint and requires the wallet's balance deltas to match the quote:

- input token decreases by no more than the max input
- output token increases by at least the quoted output minus slippage
- no other token balance decreases
- no approval is granted to a spender outside the expected set

Fails closed on a mismatch (falls through to the next candidate quote without signing).

## Why

Defence-in-depth. The existing pre-broadcast guards validate the *shape* of a swap's calldata, not its *effect* — they can't fully confirm that executing a quote does only what the user asked. This adds an effect-level check: simulate the swap and require the resulting wallet balance changes to match the quote before signing, so a quote whose outcome doesn't match intent is rejected rather than broadcast.

## How

- Deltas are computed client-side from the raw simulation result, so the check stays independent of the service that built the quote.
- Log-based delta math (Transfer/Approval + synthetic native transfers) excludes gas naturally. Primary path is `eth_simulateV1` with `traceTransfers`; falls back to `debug_traceCall` + callTracer.
- Runs after any required approval confirms on-chain, so a single-tx simulation reflects the live allowance.
- On its own gate `--no-verify-outcome`, independent of `--no-simulate`/`gasless`. Degrades with a warning when no simulation-capable endpoint is reachable rather than blocking the trade. EVM-only.

## Testing

- Unit tests (mocked RPC) for the simulator, the four assertions, and the execute-path wiring; full suite + lint green.
- Live e2e on Base: real ETH→USDC and USDC→ETH swaps (OKX + LiFi) verified end-to-end, including a real broadcast that passed verification and settled on-chain.


gulshngill:
I checked the current head and added the specific code locations below.

1. *API key on custom simulation RPC — merge blocker*
   - `src/swap-simulation.js:138-146` builds the simulation request and attaches the `apikey` header.
   - `src/rpc-urls.js:61-76` defines the trusted default and the `NANSEN_BASE_SIM_RPC` override.
   - The fix belongs in the request/header logic: only attach the Nansen key for the trusted Nansen-hosted endpoint, or enforce an explicit trusted-host check before attaching it. Add the regression test in `src/__tests__/swap-simulation.test.js` using a custom override and assert that no API key is sent.

2. *Negative exact-in output — merge blocker*
   - `src/trade-validation.js:950-974` calculates the minimum acceptable output for the exact-in path.
   - Reject `quote.outAmount`/`quote.outputAmount` when it is non-numeric or `<= 0` before calculating the minimum. Add the regression case next to the `assertSwapOutcome` tests in `src/__tests__/trade-validation.test.js`.

3. *Unexpected NFT transfers — lower-severity defense-in-depth concern*
   - `src/swap-simulation.js:94-118` parses the simulated asset changes and currently covers native/ERC-20-style deltas, but not ERC-721 or ERC-1155 transfers.
   - For a DEX-only flow, normal swaps are expected to involve native currency and ERC-20 tokens, so I am not suggesting that NFT support must be added. The concern is that arbitrary route/calldata could still transfer an NFT out of the wallet alongside an otherwise valid swap, while the outcome check would pass. Please either fail closed when an unsupported asset-transfer event is present, or explicitly document this scope and ensure the target/calldata restrictions prevent it. Add a regression test in `src/__tests__/swap-simulation.test.js` if this is intended to be covered.

The first two issues are the merge blockers.

kome12:
@gulshngill Thanks for your review. I addressed your points and also made some more improvements. Can you check again?

gulshngill:
## Fresh review findings

I re-reviewed the current head (`6c3821b40a86771d6d8d80e1f97d2a451cacddfe`). The earlier merge blockers are fixed: custom simulation RPCs no longer receive the Nansen API key, non-positive/non-numeric exact-in output is rejected, and unsupported NFT transfers are blocked.

Two findings remain:

### 1. `src/swap-simulation.js` — `SELFDESTRUCT` can double-count native value in the `debug_traceCall` fallback

`ETH_MOVING_FRAME_TYPES` still includes `SELFDESTRUCT`. Some node implementations can expose both the wallet’s top-level `CALL` and a subsequent `SELFDESTRUCT` refund frame. The accounting can then cancel or partially offset the wallet’s real native outflow, potentially allowing the balance-delta assertion to pass for a mismatched result.

Please remove `SELFDESTRUCT` from the native moving-frame set, or restrict its handling to beneficiary inflows only, with a regression test. Failing closed or under-counting is safer than accepting a potentially incorrect outcome.

### 2. `src/trading.js` — missing request intent should degrade cleanly in `verifySwapOutcome`

`verifySwapOutcome` calls `assertSwapOutcome` when `quoteData.request` is absent. For a pre-intent quote with a simulation-capable endpoint, this produces a misleading `SWAP_OUTCOME_MISMATCH` error instead of the intended re-quote/degraded behavior.

Please add an early guard after the cross-chain check:

```js
if (!quoteData?.request) {
  log('  ⚠ Swap-outcome verification skipped (no request intent — re-quote to enable it).');
  return { proceed: true };
}
```

Please add focused regression coverage for both cases.

kome12:
Thanks @gulshngill — both findings addressed in 5f2ba95.

**1. `SELFDESTRUCT` double-counting (`swap-simulation.js`)** — removed `SELFDESTRUCT` from `ETH_MOVING_FRAME_TYPES` entirely (your first option). Reasoning: the wallet is the EOA signer, so it can never `SELFDESTRUCT` itself — its real native outflow is always the top-level `CALL`. A node-surfaced `SELFDESTRUCT` refund frame to the wallet is the *only* way this could offset the outflow, and that's exactly the spurious inflow you flagged. Dropping the frame type can therefore only ever under-count a genuine selfdestruct refund into the wallet (rare, and largely neutered by EIP-6780), which makes the net native delta stricter — fail-closed, never the reverse. Regression test added: a nested `SELFDESTRUCT` refund frame to the wallet is ignored, so the full `CALL` outflow stays on the books instead of cancelling to zero.

**2. Missing request intent (`trading.js`)** — added the early guard you suggested, after the cross-chain check:

```js
if (!quoteData?.request) {
  log('  ⚠ Swap-outcome verification skipped (no request intent — re-quote to enable it).');
  return { proceed: true };
}
```

Regression test added: a pre-intent quote with a sim-capable endpoint reachable degrades to `proceed: true` without touching the sim endpoint (no misleading `SWAP_OUTCOME_MISMATCH`).

Full suite (2084 passing) + lint green.

gulshngill (dismissed):

gulshngill:
lgtm but there's a couple of suggestions from the ai reviewer.

gulshngill (dismissed):

kome12:
@gulshngill Should be ready for a review again!

gulshngill (approved):

PR #497, opened 2026-08-19
fix: reject --oid values above 2^53-1 on perp cancel

crazywriter1:
## Summary
- `perp cancel --oid` parsed via `parseInt` which silently rounds uint64 IDs above 2^53-1, potentially cancelling the wrong order
- Add `Number.isSafeInteger` guard — same logic the response path already uses to withhold unsafe oids from output
- Add regression tests: unsafe oid rejected, MAX_SAFE_INTEGER accepted

## Test plan
- [x] `npx vitest run src/__tests__/perp.test.js -t cancel`
- [x] `--oid 9007199254740992` (2^53) → error: "exceeds safe integer precision"
- [x] `--oid 9007199254740991` (2^53-1) → passes validation

### DIFF
--- a/.changeset/perp-cancel-oid-safe-integer.md
+++ b/.changeset/perp-cancel-oid-safe-integer.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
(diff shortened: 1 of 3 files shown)


crazywriter1:

kome12 (approved):

PR #498, opened 2026-08-19
fix: use pending nonce for EVM sends in transfer

crazywriter1:
## Summary
- `buildEvmTransaction` used `eth_getTransactionCount(..., 'latest')`, so mempool txs were ignored
- Back-to-back `nansen send` could reuse the same nonce (fail or silent replace)
- Fetch both `pending` and `latest`; sign at `pending`
- Stuck-nonce guard: refuse if pending−latest > 2 (same idea as `getEvmNonce` in `trading.js`)
- Error text does **not** suggest `--nonce` / `--priority-fee` (send has neither); tells the user to wait or replace from where the stuck tx was sent
- Documents the load-balanced RPC race (two calls can hit different nodes)

## Test plan
- [x] `npx vitest run src/__tests__/transfer.test.js -t EVM`
- [x] nonce calls use both `"pending"` and `"latest"`
- [x] gap > 2 throws; message includes wait + RPC caveat, not `--nonce`/`--priority-fee`

### DIFF
--- a/src/transfer.js
+++ b/src/transfer.js
@@ -120,9 +120,27 @@ async function buildEvmTransaction({ to, amount, token, privateKey, chain, max =
   const privBuf = Buffer.from(privateKey, 'hex');
   const from = deriveEvmAddress(privateKey);
 
-  // Nonce
-  const nonceHex = await rpcCall(rpcUrl, 'eth_getTransactionCount', [from, 'latest']);
-  const nonce = BigInt(nonceHex);
+  // Fetch both pending and latest nonce counts. 'pending' is used so mempool-
+  // queued transactions are counted — 'latest' alone would assign the same
+  // nonce to back-to-back sends, causing one to fail or silently replace the
+  // other. The gap check mirrors trading.js's getEvmNonce: if more than 2
+  // transactions are already queued, signing another would silently stack
+  // behind them and sit unexecutable until they clear or get replaced.
+  const MAX_PENDING_NONCE_GAP = 2;
+  const [pendingHex, latestHex] = await Promise.all([
+    rpcCall(rpcUrl, 'eth_getTransactionCount', [from, 'pending']),
+    rpcCall(rpcUrl, 'eth_getTransactionCount', [from, 'latest']),
+  ]);
+  const nonce = BigInt(pendingHex);
+  const latestNonce = BigInt(latestHex);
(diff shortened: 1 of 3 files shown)


crazywriter1:
@nansen-pr-reviewer re-review

crazywriter1:

src/transfer.js
```diff
@@ -120,9 +120,27 @@ async function buildEvmTransaction({ to, amount, token, privateKey, chain, max =
...
+  ]);
+  const nonce = BigInt(pendingHex);
+  const latestNonce = BigInt(latestHex);
+  const gap = Number(nonce - latestNonce);
+  if (gap > MAX_PENDING_NONCE_GAP) {
+    throw new Error(
+      `${from} has ${gap} unmined transactions queued on ${chain} (next mined nonce ${latestNonce}, next pending ${nonce}). ` +
+      `Signing another would queue behind them and stay unexecutable until they clear. ` +
+      `Replace the stuck transaction at nonce ${latestNonce} with a higher fee first.`,
+    );
```
kome12, line 146: Nice mirror of `getEvmNonce` from `trading.js`. One thing worth carrying over from that reference: pending and latest are read as two separate RPC calls, so against a load-balanced or public RPC they can land on different nodes and briefly report an inflated gap. This might be enough to trip this guard and refuse a perfectly valid single send when nothing is actually queued. `getEvmNonce` [documents exactly this caveat](https://github.com/nansen-ai/nansen-cli/blob/main/src/trading.js#L612) in its error text, and it's worth including here too so an operator doesn't misdiagnose from one endpoint. Very minimal change like this:
`Signing another would queue behind them and stay unexecutable until they clear. ` +
`Wait for them to clear (or replace them with a higher fee from where they were sent) before retrying. ` +
`Note that a load-balanced public RPC may report a transaction it isn't actually holding, so don't diagnose from a single endpoint.`

Above is also changing the "Replace the stuck transaction at nonce X with a higher fee first" because `nansen wallet send` doesn't expose a `--nonce` or `--priority-fee` flag

crazywriter1: Addressed in the latest commit: dropped the --nonce/--priority-fee replace hint (send has neither) and added the load-balanced RPC caveat.

kome12 (approved):

PR #499, opened 2026-08-19
API-240: Fix profiler labels — migrate addressLabels to the v1 endpoint

gulshngill:
## Summary

`NansenAPI.addressLabels()` still called a removed beta endpoint. Every `nansen profiler labels` call (and `--enrich` label lookups) hit a dead endpoint. `src/schema.json` already documented the v1 path, so code and docs disagreed.

Verified against the published OpenAPI spec (<https://docs.nansen.ai/api/profiler/address-labels>), the v1 contract differs beyond the URL:

- **Request**: top-level `{address, chain, pagination}` with `additionalProperties: false` — the old `{parameters: {address, chain}}` wrapper would be rejected, so a URL-only fix was not enough.
- **Response**: `{pagination, data: [{label, category, kind}]}` instead of the beta `labels` string array.

## Changes

- `src/api.js` — `addressLabels()` now posts to `/api/v1/profiler/address/labels` with the v1 top-level body.
- `src/cli.js` — `enrichTransfers()` maps `data[].label`, keeping the legacy `labels` fallback so existing mocks and cached responses still work.
- Tests and documentation fixtures now cover the v1 contract.
- Added a patch changeset.

## Validation

- `npm test`: 2,032 passed, 2 pre-existing skips
- `npm run lint`: clean
- Independent code review: approved, no findings

### DIFF
--- a/.changeset/fix-profiler-labels-v1-endpoint.md
+++ b/.changeset/fix-profiler-labels-v1-endpoint.md
@@ -2,4 +2,4 @@
 "nansen-cli": patch
 ---
 
-Fix `profiler labels`: call `/api/v1/profiler/address/labels` with its v1 request body — the beta endpoint previously used was removed from the Nansen API.
+Fix `profiler labels`: call `/api/v1/profiler/address/labels` with its v1 request body — the beta endpoint previously used was removed from the Nansen API. `profiler batch --include labels` now returns the label array itself instead of the raw `{pagination, data}` envelope.
--- a/.changeset/fix-profiler-labels-v1-endpoint.md
+++ b/.changeset/fix-profiler-labels-v1-endpoint.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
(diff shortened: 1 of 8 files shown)


gulshngill:
## Independent review

**Verdict: REQUEST_CHANGES**

The endpoint migration itself is correct and matches the published OpenAPI: `POST /api/v1/profiler/address/labels`, with the expected `{address, chain, pagination}` body. The full suite and lint were clean.

**Blocking finding — `src/cli.js:573`:** `batchProfile` assigns the entire response envelope to `entry.labels`. The v1 response is `{pagination, data:[...]}`, but the documented consumer shape is an array. Consumers following that shape will read `labels[0].label` and get `undefined`.

Fix:
```js
entry.labels = (await api.addressLabels({ address, chain }))?.data ?? [];
```

Also update the batch skill shape from the legacy `fullname` field to the v1 `{label, category, kind}` shape. The profiler skill’s statement that labels do not paginate is stale; please correct it as follow-up or in this PR if convenient.

PR #501, opened 2026-08-19
API-279: Guard wallet export against accidental plaintext key disclosure

gulshngill:
## Summary

`nansen wallet export <name>` previously printed raw EVM and Solana private keys to stdout with only a warning line. This PR gates plaintext disclosure behind explicit acknowledgement and adds a safer file-based export path.

- **Redacted default**: `wallet export <name>` now prints addresses + `[REDACTED]` placeholders and never decrypts — no password needed, secrets never materialize in memory on the default path.
- **`--reveal`** (explicit acknowledgement): prints plaintext keys to stdout, exactly as before. When stdout is an interactive terminal, a warning goes to stderr first (scrollback/screen-share/terminal-logging risk). Works fully non-interactively — `NANSEN_WALLET_PASSWORD=... nansen wallet export <name> --reveal` is the automation path.
- **`--file <path>`** (safer path): writes the export JSON to an exclusively-created file whose mode is pinned to exactly 0600 via `fchmod` on the open fd, independent of the process umask (a restrictive umask no longer leaves a 0400 file behind the "permissions 0600" message). Refuses to overwrite (`O_EXCL`, which also defeats pre-planted symlinks), removes the file if a mid-write/close failure leaves partial key material, and keeps stdout free of secrets. `--reveal` + `--file` together, bare `--file`, and non-string `--file` values are rejected before any decryption.
- **Parser hardening**: `reveal` joins the boolean-flag allowlist in `parseArgs` so no argument ordering can make it swallow the wallet name or bypass the `--file` conflict check.
- **Secrecy tests** (26 new): key material never appears in the redacted default, ordinary error envelopes (wrong password, missing wallet, post-decryption write failures), `DEBUG=1` stderr output, or telemetry payloads (which carry flag *names* and error codes only). Includes an ENOSPC mid-write test asserting the partial file is removed, a umask-0277 test asserting the final mode is exactly 0600, and an `fchmod` failure test asserting the empty file is removed before any key bytes are written.
- Docs: `skills/nansen-wallet-manager` documents the three modes; `skills/nansen-wallet-keychain-migration` verification steps now use `--reveal` with all output discarded, branching on the exit code (the old bare-export check would have become a silent false positive).

**Breaking-ish**: scripts that parsed `wallet export` output must add `--reveal` or switch to `--file`. Shipped as a `major` changeset with a migration note. Wallet generation, storage format, and signing are untouched.

## Explicit assumptions

- "Explicit acknowledgement" = the `--reveal` flag rather than an interactive confirm — this repo forbids interactive prompts in core, and a flag serves humans and agents identically.
- "File/clipboard" = file only; clipboard would require a new platform-specific dependency.

## Validation

- `npm test`: 73 unit files, 3038 passed, 0 failed, all mocked (latest head, run one file per process)
- `npm run lint`: clean
- Independent Codex code review: APPROVED after 5 rounds (fixed: partial-write cleanup, unlink-only-what-we-created via exclusive open, close-failure handling, no double-close)
- Security review of the diff: no findings (acknowledgement-bypass orderings, symlink/umask attacks on `--file`, error/`cause` leakage, telemetry/DEBUG surfaces all traced clean)

### DIFF
--- a/.changeset/guard-wallet-export.md
+++ b/.changeset/guard-wallet-export.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": minor
+---
+
+Guard `wallet export` against accidental plaintext key disclosure. The default output is now redacted (addresses only — no decryption, no password needed). Printing private keys to stdout requires explicit acknowledgement via `--reveal` (which also warns on stderr when stdout is an interactive terminal), and the new `--file <path>` writes keys to a file created with 0600 permissions (refusing to overwrite) while keeping stdout clean. Scripts that parsed `wallet export` output must add `--reveal` or switch to `--file`.
--- a/.changeset/guard-wallet-export.md
+++ b/.changeset/guard-wallet-export.md
@@ -1,5 +1,5 @@
 ---
-"nansen-cli": minor
+"nansen-cli": major
 ---
 
 Guard `wallet export` against accidental plaintext key disclosure. The default output is now redacted (addresses only — no decryption, no password needed). Printing private keys to stdout requires explicit acknowledgement via `--reveal` (which also warns on stderr when stdout is an interactive terminal), and the new `--file <path>` writes keys to a file created with 0600 permissions (refusing to overwrite) while keeping stdout clean. Scripts that parsed `wallet export` output must add `--reveal` or switch to `--file`.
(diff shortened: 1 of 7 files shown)


gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
@nansen-pr-reviewer re-review

PR #487, opened 2026-08-13
API-397: One-step MCP install (nansen mcp install <client>)

gulshngill:
## Summary

Implements [API-285](https://linear.app/nansen/issue/API-285) — one-step install of the hosted Nansen MCP server into local MCP clients:

```bash
nansen mcp install claude-code | claude-desktop | cursor   [--dry-run]
nansen mcp uninstall <client>
```

The command writes a `nansen` entry into the client's own config file using the API key from `nansen login` / `NANSEN_API_KEY`. No network calls, no shelling out — pure fs operations.

## Provider research (public sources)

Surveyed one-step-install mechanisms from Nansen's own MCP docs, Claude Code (`claude mcp add`, `.mcp.json`), Claude Desktop, Cursor (`~/.cursor/mcp.json` + deeplinks), VS Code (`servers` key, `code --add-mcp`), Codex, Gemini CLI, and vendor installers (Sentry wizard, Smithery CLI, Stripe, GitHub MCP badges). Key facts driving the design:

- Nansen's server is **hosted streamable HTTP** at `https://mcp.nansen.ai/ra/mcp`, auth via `NANSEN-API-KEY` header ([docs](https://docs.nansen.ai/mcp/connecting)) — so entries are remote-URL, no local server process.
- Claude Code `~/.claude.json` entries **require** `"type": "http"` next to `url`; Cursor infers from `url`; Claude Desktop's config is **stdio-only**, so it bridges via `npx mcp-remote` (pinned `mcp-remote@0.2.1`, header arg written without a space after the colon to dodge Claude Desktop's arg-splitting bug).
- Industry best practices adopted: merge-only JSON writes, backup before write, idempotent re-install (also the key-rotation path), `--dry-run`, uninstall support.

## Security decisions (threat-modeled independently)

- **No config clobbering** (highest practical risk — these files hold users' *other* MCP servers): only `mcpServers.nansen` is ever assigned; all sibling servers and unrelated keys pass through. Unparseable JSON → refuse with the file path, never repair/overwrite. `.bak` copy before every install write. Atomic temp-file + rename so a crash can't truncate the config. Type guard on `mcpServers`.
- **Key exposure**: the key is never printed — not in output, `--dry-run` (redacted), or errors. New dirs 0700, files 0600, backup 0600, existing target chmod'd 0600 post-write (best-effort). Explicit plaintext + settings-sync warnings on install. Telemetry already sends flag *names* only, so no key material can leak there.
[...]

PR #502, opened 2026-08-19
API-285: Authenticated MCP setup verification (nansen mcp verify)

gulshngill:
## Summary

Adds authenticated verification for the MCP setup.

```bash
nansen mcp verify            # checks the saved credential
nansen mcp verify <client>   # checks the credential installed in that client's config
```

## Why

Some MCP protocol discovery operations can succeed without credentials, so a typo, rotated, or placeholder key may appear to install successfully and only fail on the first real data call. `verify` makes that call up front and maps failures to an actionable next step.

## Design / security

- Client mode verifies the credential in the client's configured server entry. It accepts hand-written entries that still target the official server URL/transport, warns about non-security-relevant differences, and refuses entries that would send the key elsewhere before making a network call.
- Redirects are rejected so the credential cannot be forwarded to another URL.
- Credentials are never printed; response bodies are redacted before being used in messages.
- `--dry-run` is rejected up front so the credit-consuming path is never reached silently; credit cost is disclosed in output, usage, schema, and README.
- Observed failure shapes map to actionable guidance, including missing credentials, invalid credentials, HTTP errors, malformed responses, connectivity failures, and timeouts.
- The transport uses stateless streamable HTTP with the required content negotiation and response parsing.
- `install` now points users to `verify` as the final setup step.

## Tests

- `npm test`: 54 files, 2,010 passed / 2 skipped
- `npm run lint`: clean
- Coverage includes request shape, redirects, response classification, SSE response matching, status-family mapping, config-drift refusal, dry-run rejection, credential redaction, and CLI registration.


gulshngill:
## Independent review

**Verdict: REQUEST_CHANGES**

Protocol behavior, credential redaction, endpoint pinning, live verification, and the existing test suite were checked and are generally sound. Two fixes are needed:

1. **Medium — `src/commands/mcp.js:104-112`, used at `376-379`:** `extractInstalledKey` requires deep equality with the exact config produced by the installer. Valid manually-created or deep-link configs with the official URL/header but extra or reordered fields are rejected, even though the verification request uses the module’s constant official URL and never trusts the config URL. Gate on the security-relevant fields instead: official URL (or the pinned official desktop transport) plus a non-empty key; warn about extra fields rather than treating them as drift.

2. **Medium — missing handler-level failure coverage around `src/commands/mcp.js:437-440`:** the primary live failure shape is HTTP 200 with `result.isError` and an invalid-key message. Classification helpers are tested, but the command path is not. Add a test that drives this response through the command and asserts exit 1/actionable output without echoing the key. The current coverage claim that all classification branches are covered is not accurate.

Follow-ups noted by the reviewer: classify insufficient credits separately, verify the installed desktop transport rather than only direct HTTP, and tighten the success gate. These are non-blocking relative to the two fixes above.

gulshngill:
Closing as a duplicate of #508 — both implement `nansen mcp verify`, and they cannot coexist: same command in `src/schema.json` and `src/cli.js`, and both claim `.changeset/mcp-verify.md`.

**Why #508 wins**

- **It matches API-318's acceptance criteria and this PR cannot.** That ticket asks for a verification path that is *client-independent* and "requires no repository checkout or internal script". #508 is built to that spec; this PR is deliberately client-*aware*, which is a different product.
- **Mergeability.** #508 is based on `main` and is `MERGEABLE`. This PR is stacked on `feat/api-285-mcp-install` (#487) and is `CONFLICTING`/`DIRTY`, untouched since 2026-08-19.
- **Failure taxonomy.** #508 separates auth rejection, rate limiting, and insufficient credits, and orders the checks so the most specific cause wins. This PR's `classifyVerifyResult` is coarser.

**Not lost — worth porting on top of #508 once #487 lands**

Two things here have no equivalent in #508, and both are genuinely good:

- `extractInstalledKey(client, entry)` — pulls the key out of a client's config, with hardened validation (refuses non-official URLs, `npx` overrides, registry redirects).
- `entryDriftNotes(client, entry)` — diffs an installed entry against what `install` would write, field by field, and reports drift.

Neither makes sense without `nansen mcp install`, so they belong *after* #487 anyway. #508 verifies a key; these verify a **config**. They compose rather than compete.

**The branch is not deleted** — `feat/api-285-mcp-verify` stays, so this is recoverable and portable.

**Knock-on: #507 needs a new base.** API-320 is stacked on this branch, so it now points at something that will never merge. It needs rebasing onto `main` (or onto #487 once that lands).

One correction carried over: this PR's `extractInstalledKey` requires `MCP_REMOTE_PIN` to match exactly, and #487's pin has since moved from `mcp-remote@0.1.38` to `0.2.1` to match the published docs. Any port must use the current pin, or it will reject configs that follow the documentation.

PR #507, opened 2026-08-19
API-320: Standardize MCP client credential handling (opt-in --env-ref, no key echoes)

gulshngill:
Stacked on #502 (which stacks on #487) — review only the top commit.

## What

Defines the safest practical credential path per supported MCP client and closes the remaining key-leak paths in generated configs, backups, failure messages, and shell history.

### Opt-in `nansen mcp install <client> --env-ref`

Writes an environment-variable **reference** instead of the key literal, so the generated config never holds the key. No login required — the CLI never reads the key in this mode.

| Client | Default (unchanged) | `--env-ref` | Ceiling |
| --- | --- | --- | --- |
| claude-code | key literal in `headers` | `"NANSEN-API-KEY": "${NANSEN_API_KEY}"` | some client versions send the literal reference in HTTP headers |
| cursor | key literal in `headers` | `"NANSEN-API-KEY": "${env:NANSEN_API_KEY}"` | remote-header interpolation has known gaps; stdio is reliable |
| claude-desktop | key in config `env` block (argv already env-referenced) | `env` block omitted; the pinned mcp-remote bridge inherits `NANSEN_API_KEY` from the OS environment | no native config expansion — OS-level env vars required |

Indirection is **opt-in, not the default**, because both claude-code and cursor have documented-but-buggy header expansion: a config the client fails to expand silently sends the literal `${...}` string as the API key. The default inline install now points at `--env-ref` in its plaintext note; README documents each pattern and its ceiling.

### verify understands references

`nansen mcp verify <client>` resolves the exact reference forms from the current shell and fails actionably (without echoing anything) when `NANSEN_API_KEY` is unset. Near-miss references — any `${...}` value that isn't the exact supported form — are refused, because the client could expand them to a credential verify never tested. Scope is documented honestly: verify proves config shape + current-shell resolution, not that the client itself expands the reference.

### No more key copies in backups

Replacing an existing nansen entry used to byte-copy the old config (key included) into `<config>.bak`, so an `--env-ref` migration stranded the old key on disk. The backup now redacts the nansen entry's credential slots (headers in any casing, desktop `env` block) while preserving every other entry verbatim; the nansen entry is regenerable via re-install.

### Login hardening

`nansen login` help now leads with history-safe methods (`--human`, secret-manager command substitution) and states plainly that literal `--api-key` values and inline `NANSEN_API_KEY=<key>` assignments are recorded in shell history. The login verification failure message redacts the key before relaying upstream error text.

## Tests

- `npm test`: 2016 passed, 2 skipped (54 files); `npm run lint` clean.
- New coverage: env-ref entry shapes per client; reference resolution / near-miss refusal / empty-env-as-unset; drift notes clean on env-ref entries; logged-out `--env-ref` install in both flag positions; verify resolving from env and rejecting unset refs without a network call; backup redaction on migration; login help ordering and failure-message redaction; no-echo assertions throughout.
- Live smoke against the hosted MCP server: `--env-ref` install logged-out → config holds only the reference; `verify` with the variable unset → actionable error; with it set → authenticated data call succeeds.

### DIFF
--- a/.changeset/mcp-env-ref.md
+++ b/.changeset/mcp-env-ref.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": minor
+---
+
+Add opt-in `nansen mcp install --env-ref` credential references and harden login guidance against shell-history and error-message credential leaks.
--- a/README.md
+++ b/README.md
@@ -61,13 +61,26 @@ nansen mcp install claude-code       # ~/.claude.json (user scope)
 nansen mcp install claude-desktop    # macOS/Windows only; bridges via pinned mcp-remote
 nansen mcp install cursor            # ~/.cursor/mcp.json
 nansen mcp install cursor --dry-run  # print what would be written (key redacted)
+nansen mcp install cursor --env-ref  # reference NANSEN_API_KEY instead of storing it
 nansen mcp uninstall <client>        # remove the entry (add --dry-run to preview)
-nansen mcp verify [client]           # prove the setup with one real authenticated data call
+nansen mcp verify [client]           # check setup with one real authenticated data call
 ```
 
-Uses the API key from `nansen login` / `NANSEN_API_KEY`; re-run `install` after rotating your key. Installs are merge-only and atomic: existing servers and settings are preserved, a `.bak` copy is written first, and the CLI refuses to touch a config it can't parse. Note the client config stores the API key in plaintext — new files are created with `0600` permissions. Restart the client after installing. For other clients, see [docs.nansen.ai/mcp/connecting](https://docs.nansen.ai/mcp/connecting).
+Uses the API key from `nansen login` / `NANSEN_API_KEY`; re-run `install` after rotating your key. Installs are merge-only and atomic: existing servers and settings are preserved, a `.bak` copy is written first, and the CLI refuses to touch a config it can't parse. Note the client config stores the API key in plaintext by default (see `--env-ref`) — new files are created with `0600` permissions. Restart the client after installing. For other clients, see [docs.nansen.ai/mcp/connecting](https://docs.nansen.ai/mcp/connecting).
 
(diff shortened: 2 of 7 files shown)


gulshngill:
Heads-up: **this PR's base branch is now orphaned.** #502 was closed as a duplicate of #508, and this PR is based on `feat/api-285-mcp-verify` — #502's branch. That branch is preserved, but it will never merge, so this PR needs a new base.

Retarget onto `main` (or onto `feat/api-285-mcp-install` if you want it stacked behind #487, which is now rebased and `MERGEABLE`).

While rebasing, two things changed under this PR:

1. **`MCP_REMOTE_PIN` moved `mcp-remote@0.1.38` → `0.2.1`** in #487, matching what the public docs and the nansen-cli README now publish. The old pin is ~7 months and ~13 releases behind.
2. **The `--env-ref` premise is worth re-checking.** This PR makes env indirection opt-in, but #487 already writes the `env` form by default for `claude-desktop` (`args: [..., 'NANSEN-API-KEY:${NANSEN_API_KEY}'], env: { NANSEN_API_KEY: apiKey }`), and the published README documents that as the recommended shape. So for that client the flag may be redundant. The clients where indirection is still a real choice are `claude-code` and `cursor`, which use native `url` + `headers` rather than a spawned process.

gulshngill:
Closing after a four-way review (code, functional, integration, ticket-fit). Not because the work is bad — the hygiene pieces are good and are being re-landed — but because the PR is structurally unsalvageable and one feature is unsafe:

1. **The base is dead.** This stacks on #502's branch; #502 was closed in favour of #508 (client-independent verify). #507's own commits modify ~360 lines of #502-only verify machinery (`extractInstalledCredential`, drift notes, verify gating, even `redactSecret`'s definition site). A rebase onto #487 hits 10 conflict hunks, most of which resolve by deleting the code being rebased — and commit 54bce66 cannot be cherry-picked alone because it refines lines c52a3b0 created.
2. **cursor/claude-code `--env-ref` is broken by design.** It writes `${env:NANSEN_API_KEY}` / `${NANSEN_API_KEY}` into native HTTP `headers` — no process is spawned, so nothing expands the reference. Verified live: the literal reference sent as a header → 401 inside a 200. Worse, `verify` resolves the reference in the CLI's own shell and prints the same "✓ Authenticated" as a real success — a false green light for a config the client cannot use. Only claude-desktop's env-ref (spawned mcp-remote expands it) is mechanically sound, and that was verified working end to end.

**Re-landed / to re-land:**
- Login key-echo hygiene (main still relays upstream error text containing the key; history-safe login help) → fresh PR on `main`, with `redactSecret` housed outside mcp.js.
- Backup redaction (`redactNansenEntryCredential` + the 54bce66 refinements — reviewed working) and **claude-desktop-only** `--env-ref` → follow-up PR once #487 and #508 land.
- Dropped: cursor/claude-code env-ref (above), all verify gating (dead with #502).

Branch `feat/api-320-credential-handling` is preserved — the redaction implementation and its tests are copyable nearly verbatim.

PR #508, opened 2026-08-19
API-318: Add `nansen mcp verify` — client-independent authenticated MCP setup verification

gulshngill:
## Problem

MCP clients can list the hosted Nansen MCP server's tools **without a valid API key** — the server serves `tools/list` (and free tools) unauthenticated, and tool-argument validation also runs before auth. So a broken credential setup looks successful until the first paid data call fails. There was no user-facing way to verify an MCP setup end to end.

## Solution

New diagnostic command — client-independent, runnable with zero checkout:

```bash
npx -y nansen-cli mcp verify --api-key <key>
```

It performs a real authenticated data-path check, not just discovery:

1. **`mcp-api-key`** — resolves the key (`--api-key` → `NANSEN_API_KEY` → `~/.nansen/config.json`) and reports the masked key + source.
2. **`mcp-server`** — JSON-RPC `tools/list` for reachability (explicitly labelled "unauthenticated; reachability only").
3. **`mcp-auth`** — `tools/call` on the cheapest paid canary tool (`nansen_score_top_tokens`, ~1 credit). Server error text is mapped to actionable next steps: key rejected → check the client's `NANSEN-API-KEY` header / rotate at app.nansen.ai; insufficient credits → top up; rate limited → retry (and the setup is *not* reported verified).

Verified ⟺ the paid canary call succeeds with a well-formed result. On failure the command exits non-zero through the CLI's unified error envelope (`code: MCP_VERIFY_FAILED`); `--json` returns the structured checks (failure carries them in `details`). Success message is honest about scope: the CLI proves the *key* works against the server's paid data path and reminds the user to put that same key in their client's header; README documents an in-client canary prompt for that final hop.

No new dependencies — the stateless streamable-HTTP server needs exactly two `fetch` POSTs (JSON or SSE-framed responses both handled).

## Changes

- `src/mcp-verify.js` (new) — MCP JSON-RPC probe + checks + report formatting
- `src/cli.js` — `mcp verify` wiring, HELP entry; bare `nansen mcp` prints usage only (never runs the paid probe implicitly)
- `src/doctor.js` — export `resolveAuthConfig`, extract shared `formatChecks` (doctor output unchanged, proven by existing tests)
- `src/schema.json` — `mcp verify` command schema
- `README.md` — "Verify your MCP setup" section + troubleshooting row
- `src/__tests__/mcp-verify.test.js` (new, 22 tests) — mocked-fetch coverage of SSE/JSON parsing, every auth-failure mapping, timeout/network/unparseable paths, key precedence, report format, and the non-zero `--json` exit contract
- `.changeset/mcp-verify.md` — minor

## Testing

- `npm test`: 2310 passed, 2 skipped (57 files) — all fetch mocked. Re-run after merging `main` (2026-08-28).
- `npm run lint`: clean
- Live smoke against production MCP: bogus key → key-rejected mapping + exit 1; valid key → verified + exit 0; `--json` failure → single machine-readable envelope; bare `nansen mcp` → usage, no network call

Note for docs: the verify one-liner is worth adding to the MCP "Connecting" page on docs.nansen.ai (external repo, out of scope here).

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -16,6 +16,7 @@ import { resolveAddress, isEnsName } from './ens.js';
 import fs from 'fs';
 import { getUpdateNotification, getUpgradeNotice, scheduleUpdateCheck } from './update-check.js';
 import { getAuthStatus, runDoctorChecks, runConnectivityChecks, formatDoctorReport } from './doctor.js';
+import { DEFAULT_MCP_URL, formatMcpVerifyReport, runMcpVerifyChecks } from './mcp-verify.js';
 import { refreshCostMapIfStale, getCostForEndpoint, creditsCharged } from './cost-cache.js';
 import { creditWarning, noticeWarnings } from './response-meta.js';
 import { trackCommandSucceeded, trackCommandFailed } from './telemetry.js';
@@ -739,6 +740,7 @@ COMMANDS:
   login       Save API key (--api-key <key>, --human, or NANSEN_API_KEY env var)
   logout      Remove saved API key
   doctor      Diagnostics: auth, wallets, caches, connectivity (--offline --json)
+  mcp         Verify hosted MCP setup with an authenticated data call (~1 credit)
   schema      JSON schema for all commands (use "nansen schema <cmd>" for one)
   cache       clear
   changelog   --since <version> to filter
@@ -886,7 +888,9 @@ export function buildCommands(deps = {}) {
     deleteConfigFn = deleteConfig,
     getConfigFileFn = getConfigFile,
     isTTY = process.stdin.isTTY,
-    env = process.env
+    env = process.env,
+    fetchFn = fetch,
+    devConfigPath
   } = deps;
 
   const cmds = {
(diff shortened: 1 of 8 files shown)


gulshngill:
This is the chosen `nansen mcp verify` implementation — #502 has been closed as a duplicate.

Rationale recorded on #502, in short: this PR matches API-318's acceptance criteria (client-independent, no repository checkout or internal script), it is based on `main` and `MERGEABLE`, and its failure taxonomy separates auth rejection, rate limiting, and insufficient credits rather than collapsing them.

**Only thing blocking it is a review.** CI is green.

Two items to port on top later, from #502's preserved branch (`feat/api-285-mcp-verify`), once #487 lands — they verify a *config* rather than a *key*, so they compose with this rather than compete:

- `extractInstalledKey(client, entry)` — extract the key from a client config, refusing non-official URLs, `npx` overrides, and registry redirects.
- `entryDriftNotes(client, entry)` — field-by-field diff of an installed entry against what `install` writes.

Any port must use the current `MCP_REMOTE_PIN` (`mcp-remote@0.2.1`), not the `0.1.38` those functions were written against, or it will reject configs that follow the published docs.

src/mcp-verify.js
```diff
@@ -0,0 +1,285 @@
...
+}
+
+function authFailureCheck(message, httpStatus = null) {
+  const text = responseText(message);
+  const codes = [httpStatus, message?.error?.code, message?.result?.code].filter(Boolean).join(' ');
+  const combined = `${codes} ${text}`;
+
+  // Order matters: auth rejection first, then rate limit before credits —
+  // rate-limit texts often mention "credit rate limit" and must stay a warn.
+  if (/\b40[13]\b|unauthorized|forbidden|api[- ]key.*(?:required|invalid|reject)|header is required/i.test(combined)) {
```
Codier: [P1] HTTP 403 is misclassified as a rejected API key

This regex branch matches any 403 (or "forbidden") before the credits check ever runs, so a 403 with "insufficient credits" — which Nansen's own gateway maps to CREDITS_EXHAUSTED (src/api.js statusToErrorCode: 403 + "insufficient"/"credit") — is reported as "MCP server rejected the API key" with a create/rotate fix. A valid-key user with zero balance is sent to the key management page, and on plans that cap the number of keys the suggested create/rotate can itself fail. Move the credits test before the generic 40x match, or exclude a bare 403 from the rejection branch.

src/cli.js
```diff
@@ -919,6 +923,59 @@ export function buildCommands(deps = {}) {
...
+        log(formatMcpVerifyReport(checks, url, true));
+        return;
+      }
+
+      const reason = (checks.find(checkItem => checkItem.status === 'error')
+        || checks.find(checkItem => checkItem.id === 'mcp-auth' && checkItem.status !== 'ok'))?.message
+        || 'the paid data path did not complete';
+      const message = `MCP setup verification failed — ${reason}`;
+      if (flags.json) throw new CommandError(message, 'MCP_VERIFY_FAILED', result);
+      log(formatMcpVerifyReport(checks, url, false));
```
Codier, line 983: [P2] Non-JSON failure double-outputs report and error envelope

log(formatMcpVerifyReport(...)) prints the full human report, then the throw makes runCLI serialize a {"success":false,"code":"MCP_VERIFY_FAILED",...} envelope to the same stdout (details omitted). Every failed default-mode verify therefore emits two different output shapes, a report followed by a JSON envelope, while the success path prints only a report and the --json failure path prints only an envelope. Have this path exit non-zero without re-emitting the envelope.

src/cli.js
```diff
@@ -919,6 +923,59 @@ export function buildCommands(deps = {}) {
...
+      if (flags['api-key']) {
+        throw new NansenError('--api-key requires a value. Usage: nansen mcp verify --api-key <key>', ErrorCode.MISSING_PARAM);
+      }
+      // parseArgs JSON-parses option values, so `--api-key null` arrives as
+      // null and a repeated flag as an array — both must fail, not fall back.
+      if ('api-key' in options && typeof options['api-key'] !== 'string') {
+        throw new NansenError('--api-key must be a single key string. Usage: nansen mcp verify --api-key <key>', ErrorCode.INVALID_PARAMS);
+      }
+
+      const url = options.url || DEFAULT_MCP_URL;
```
Codier, line 955: [P3] --url is never type-validated

parseArgs turns repeated string options into arrays, and --api-key was guarded against that, but url is not: repeated --url flows an array into fetch and surfaces as "MCP server unreachable: ... https://a,https://b" instead of a validation error, and --url null silently falls back to the default. Reject non-string url values, mirroring the api-key guard.

Codier (commented):
Thorough review posted as three inline comments. All findings independently reproduced at this head; suite (2311 passed, 2 skipped) and lint are clean.

src/mcp-verify.js
gulshngill: Fixed in babb508. Reordered most-specific first: rate limit → credits → auth (the branch comment claiming auth-first was the bug's own justification). The credits branch now also matches bare "credit(s)" per the gateway's CREDITS_EXHAUSTED mapping, and "credit rate limit" stays a warn because the rate-limit branch runs first. Tests: 403+"insufficient credits" → credits error with top-up fix; bare 403 Forbidden → still key rejection; "credit rate limit" → still a warn. The first test was run against the unfixed code and fails.

src/cli.js
gulshngill: Fixed in babb508. The command marks the thrown error `reported = true` after printing the report; runCLI's catch skips the envelope for such errors but still calls trackFailed and exits 1. --json is unchanged (envelope only, no report). Verified live against prod: default-mode bad key prints the report once, exit 1, zero envelope on stdout; --json emits exactly the envelope.

src/cli.js
gulshngill: Fixed in babb508. `--url` now has the exact same three guards as `--api-key`: valueless flag → MISSING_PARAM, repeated/JSON-parsed non-string (array, null) → INVALID_PARAMS, all before any fetch. Tests assert the mock fetch is never called on the rejected forms.

gulshngill:
Thanks for the review @Codier could you help have a look at it again

Codier (approved):
Verified the fixes on the new head — checks, failure output, and url guards all behave as intended; tests and lint green. LGTM.

PR #511, opened 2026-08-21
Add nansen mcp install/verify for local MCP clients (Cursor first)

gulshngill:
## What

Adds a real one-command MCP setup to the CLI — the command the docs always wanted to exist (see nansen-api#1803, API-317):

```bash
nansen login --api-key <key>
nansen mcp install cursor
nansen mcp verify cursor
```

- **`nansen mcp install cursor`** merges a native `url` + `headers` Nansen entry into `~/.cursor/mcp.json` (or `--config-path`), using the saved login key, `NANSEN_API_KEY`, or `--api-key`. It preserves other configured servers, refuses to clobber a file that isn't valid JSON, and chmods the file to `0600` since it embeds a credential.
- **`nansen mcp verify cursor`** proves the setup in two steps: a streamable-HTTP `initialize` handshake against the configured URL (transport), then a credit-free account check with the configured key (auth). The second step exists because the MCP server accepts `initialize` and `tools/list` unauthenticated — a handshake alone never proves the key works.

The client registry is a map, so adding e.g. `claude-desktop` later is one entry.

## Why

API-317 flagged that the documented Cursor "one-click" deep link shipped a placeholder key inside an opaque base64 payload. nansen-api#1803 replaces it with copy-paste JSON; this PR gives users the actual one-command path and a machine-checkable verify step.

## Testing

- 21 new unit tests (`src/__tests__/mcp.test.js`); full suite passes (2115 passed, 2 skipped), eslint clean.
- Verified live against `mcp.nansen.ai`:
  - happy path with a valid key → `transport: ok`, `auth: ok`, server `nansen-mcp 3.2.4`, exit 0
  - invalid key → handshake passes, auth step fails with structured `INVALID_API_KEY` + resolution steps, exit 1
  - merge behavior, invalid-JSON refusal, and `0600` permissions confirmed on disk

## Follow-up

Once released, nansen-api's `gitbook/mcp/connecting.md` Cursor section can add these commands back (they were removed from #1803 because they didn't exist yet). README here already documents them.

### DIFF
--- a/.changeset/mcp-install-command.md
+++ b/.changeset/mcp-install-command.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": minor
+---
+
+Add `nansen mcp install <client>` and `nansen mcp verify <client>` (Cursor first). `install` merges a native url+headers Nansen entry into the client's `mcp.json` using the saved login key (or `--api-key`), preserves other configured servers, refuses to clobber invalid JSON, and restricts the file to owner-only permissions. `verify` proves the endpoint with a streamable-HTTP initialize handshake and — because the handshake succeeds even unauthenticated — validates the configured key with a credit-free account check.
--- a/README.md
+++ b/README.md
@@ -59,7 +59,14 @@ Connect any MCP client to Nansen's streamable HTTP server:
 - **Authentication:** `NANSEN-API-KEY` header
 - **API key:** [app.nansen.ai/auth/agent-setup](https://app.nansen.ai/auth/agent-setup)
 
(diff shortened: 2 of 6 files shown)


gulshngill:
Closing as a duplicate: this was opened without checking the existing in-flight MCP stack, which is more complete and pre-dates it — #487 (`nansen mcp install` for claude-code/claude-desktop/cursor, with atomic merge-writes, backups, --dry-run, uninstall, key redaction) and #502/#508 (`nansen mcp verify` via a real authenticated MCP tools/call, which is a stronger auth proof than the account-endpoint check here). One useful data point from this PR's live testing carries over: the MCP server accepts initialize and tools/list unauthenticated, so any verify must make a real data call — both #502 and #508 already do this.

PR #512, opened 2026-08-21
Bind Solana swap execution to the persisted request intent

kome12:
## Summary
- Wire the existing intent-binding checks (`assertQuoteMatchesRequest`, a new `assertCompleteSolanaRequestIntent`) into all three Solana signing paths (local, Privy, WalletConnect), closing the gap where Solana previously signed the aggregator's transaction verbatim with no validation. This is the Solana counterpart to PR #494's EVM hardening.
- `--swap-mode exactOut` now requires `--max-input` on Solana too (previously EVM-only). A cap auto-derived from the API's own quote response can't independently guard against an inflated one — an explicit, user-supplied ceiling is required on every chain now, matching EVM.

## Known follow-up (not in scope here)
While doing live e2e testing, found a **pre-existing bug on `main`**, unrelated to this diff: Relay-sourced Solana bridge quotes return raw, uncompiled instructions (`{instructions, addressLookupTableAddresses}`) rather than a ready-to-sign transaction, and the signing code has no branch for that shape — crashes with `ALL_QUOTES_FAILED`. Filed separately as API-353; confirmed via `git diff origin/main` that the affected code is untouched by this PR.

Two smaller/cosmetic review findings also left as follow-up (not correctness bugs): duplicated `assertCompleteEvmRequestIntent`/`assertCompleteSolanaRequestIntent` (copy-paste instead of a shared helper), and duplicated native-SOL-mint literals across `trading.js`/`trade-validation.js` (to avoid a circular import).

## Test plan
- [x] `npm test` — 2105 passed, 2 skipped
- [x] `npm run lint` — clean
- [x] Live mainnet e2e (`bridge-test` wallet): SOL ↔ USDC same-chain round-trip via the local-wallet signing path this PR changes — both legs succeeded, net cost ~143,584 lamports (fees/slippage only)
- [x] Code review pass; one confirmed correctness bug found and fixed (the self-referential `maxInputAmount` cap), see commit c1b8ffb


gulshngill:
Follow-up review findings:

1. `src/trading.js:2053-2056` and `:2277-2279` — The comments overstate the protection provided by the new validation. The implementation validates persisted request/quote metadata, but does not inspect the serialized Solana transaction instructions. Please narrow the wording so it does not imply that the signed transaction’s contents are fully bound to the persisted intent.

2. `src/trading.js:2274` — The local Solana signing path does not apply the same fail-closed wallet-address check used by the Privy path. Please add the equivalent guard so all signing paths consistently reject an unavailable signer address before intent validation/signing.

3. `src/__tests__/trading.test.js:1580-1731` and `:4557-4737` — There is some redundant coverage between the Solana validation suites. Please consolidate the overlapping cases while retaining the distinct signer-path coverage.

4. Please rebase the branch onto the current base branch and remove any duplicate hunk that is already present upstream before merging.

kome12:
> Follow-up review findings:
> 
> 1. `src/trading.js:2053-2056` and `:2277-2279` — The comments overstate the protection provided by the new validation. The implementation validates persisted request/quote metadata, but does not inspect the serialized Solana transaction instructions. Please narrow the wording so it does not imply that the signed transaction’s contents are fully bound to the persisted intent.
> 2. `src/trading.js:2274` — The local Solana signing path does not apply the same fail-closed wallet-address check used by the Privy path. Please add the equivalent guard so all signing paths consistently reject an unavailable signer address before intent validation/signing.
> 3. `src/__tests__/trading.test.js:1580-1731` and `:4557-4737` — There is some redundant coverage between the Solana validation suites. Please consolidate the overlapping cases while retaining the distinct signer-path coverage.
> 4. Please rebase the branch onto the current base branch and remove any duplicate hunk that is already present upstream before merging.

@gulshngill Thanks addressed your findings. Rebased from the latest main as well.

In terms of scope, this PR only addresses the metadata binding and not the transaction instructions. Follow up PRs will address this.

gulshngill (approved):

PR #513, opened 2026-08-21
Fix Solana execute crash on Relay's raw-instruction bridge quotes

kome12:
## Summary

- `trade execute` normalized Solana quote transactions into two known shapes — Jupiter (base64 string) and OKX (`{data: base58}`) — but Relay's Solana-source bridge quotes return a third, uncompiled shape (`{instructions, addressLookupTableAddresses}`). That shape fell through and crashed on `Buffer.from(object, 'base64')`, so any Solana→EVM bridge failed whenever Relay won the aggregator price race, even though `trade quote` reported success.
- Adds a compiler that turns Relay's raw instructions into a signable `VersionedTransaction`, keeping all accounts static (address lookup tables are a size optimization, not a correctness requirement — skipped for now, with a clear error if a future quote is ever too large to fit without them).
- No changes to the signing paths themselves — this is purely a normalization step, so Jupiter/OKX behavior is unchanged.

## Test plan

- [x] `npm test` — all existing tests pass, plus new unit tests covering: Jupiter/OKX passthrough (no RPC call), compiling a real captured Relay-shaped quote into a valid, signable transaction (byte-level assertions), an oversized-transaction error path, and a missing-signer error path.
- [x] `npm run lint`
- [x] Live e2e: executed a real Solana→Base bridge via a forced Relay quote (`--aggregator relay`). Confirmed the exact previously-crashing transaction shape, and it signed + broadcast successfully. Verified the destination balance increased by the quoted amount (bridge completed).

### DIFF
--- a/.changeset/fix-solana-relay-bridge-signing.md
+++ b/.changeset/fix-solana-relay-bridge-signing.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
+Fix `trade execute` crashing on Solana-source bridge quotes from the Relay aggregator, which return raw uncompiled instructions instead of a ready-to-sign transaction. These are now compiled client-side before signing.
--- a/src/__tests__/trading.test.js
+++ b/src/__tests__/trading.test.js
@@ -43,6 +43,7 @@ import {
   simulateEvmCall,
   verifySwapOutcome,
   verifySolanaSwapOutcome,
+  compileRawSolanaTransaction,
   getBridgeStatus,
   pollBridgeStatus,
   saveTxRecord,
@@ -6475,5 +6476,22 @@ describe('Relay Solana-source bridge: raw-instruction transaction shape', () => 
 
     delete process.env.NANSEN_WALLET_PASSWORD;
     vi.unstubAllGlobals();
+  });
+
+  it('rejects an instruction with malformed hex data instead of silently truncating it', async () => {
+    const signer = generateSolanaWallet().address;
+    const badQuote = (data) => ({
+      instructions: [{
+        keys: [{ pubkey: signer, isSigner: true, isWritable: true }],
+        programId: generateSolanaWallet().address,
+        data,
+      }],
+    });
+    // Odd-length and non-hex both drop/mangle bytes under Buffer.from(_, 'hex').
+    // Throws during instruction decode, before any RPC blockhash fetch.
+    await expect(compileRawSolanaTransaction(badQuote('abc'), 'http://unused', async () => signer))
+      .rejects.toThrow(/not valid hex/);
+    await expect(compileRawSolanaTransaction(badQuote('deadXY'), 'http://unused', async () => signer))
+      .rejects.toThrow(/not valid hex/);
   });
 });
--- a/src/__tests__/trading.test.js
+++ b/src/__tests__/trading.test.js
@@ -44,6 +44,7 @@ import {
   verifySwapOutcome,
   verifySolanaSwapOutcome,
   compileRawSolanaTransaction,
+  normalizeSolanaTransaction,
   getBridgeStatus,
   pollBridgeStatus,
   saveTxRecord,
@@ -6493,5 +6494,26 @@ describe('Relay Solana-source bridge: raw-instruction transaction shape', () => 
       .rejects.toThrow(/not valid hex/);
     await expect(compileRawSolanaTransaction(badQuote('deadXY'), 'http://unused', async () => signer))
       .rejects.toThrow(/not valid hex/);
+  });
+
+  it('normalize dispatches the raw-instructions shape even when a data field is also present', async () => {
+    const signer = generateSolanaWallet().address;
+    // A hypothetical future shape carrying BOTH instructions and data: the
+    // Relay compiler must win, not the OKX base58-decode branch. base58Decode
+    // would throw on this non-base58 data, so reaching it at all is the failure.
+    vi.stubGlobal('fetch', vi.fn().mockResolvedValue({
+      json: () => Promise.resolve({ result: { value: { blockhash: generateSolanaWallet().address } } }),
+    }));
+    const mixed = {
+      data: 'not-base58-@@@',
+      instructions: [{
+        keys: [{ pubkey: signer, isSigner: true, isWritable: true }],
+        programId: generateSolanaWallet().address,
+        data: 'deadbeef',
+      }],
+    };
+    const b64 = await normalizeSolanaTransaction(mixed, 'http://unused', async () => signer);
+    expect(Buffer.from(b64, 'base64').length).toBeGreaterThan(0);
+    vi.unstubAllGlobals();
   });
 });
--- a/src/__tests__/trading.test.js
+++ b/src/__tests__/trading.test.js
@@ -6516,4 +6516,41 @@ describe('Relay Solana-source bridge: raw-instruction transaction shape', () => 
     expect(Buffer.from(b64, 'base64').length).toBeGreaterThan(0);
     vi.unstubAllGlobals();
   });
+
+  it('rejects an instruction set that requires more than one signature', async () => {
+    const signer = generateSolanaWallet().address;
+    const coSigner = generateSolanaWallet().address;
+    // Two distinct declared signers: only the wallet's own single signature can
+    // ever be provided (slot 0), so this must fail closed before any RPC call
+    // rather than leave the co-signer's slot empty and fail opaquely on-chain.
+    const twoSigner = {
+      instructions: [{
+        keys: [
+          { pubkey: signer, isSigner: true, isWritable: true },
+          { pubkey: coSigner, isSigner: true, isWritable: true },
+        ],
+        programId: generateSolanaWallet().address,
+        data: 'deadbeef',
+      }],
+    };
+    await expect(compileRawSolanaTransaction(twoSigner, 'http://unused', async () => signer))
+      .rejects.toThrow(/requires 2 signatures/);
+  });
+
+  it('rejects a transaction too large to compile without address lookup tables', async () => {
+    const signer = generateSolanaWallet().address;
+    // A single instruction whose data alone blows past Solana's 1232-byte packet
+    // limit: skipping ALTs is only valid while the static tx still fits, so an
+    // oversized route must throw, not silently build an unsignable transaction.
+    const bigData = 'ab'.repeat(1300); // 1300 bytes, valid hex
+    const oversized = {
+      instructions: [{
+        keys: [{ pubkey: signer, isSigner: true, isWritable: true }],
+        programId: generateSolanaWallet().address,
+        data: bigData,
+      }],
+    };
+    await expect(compileRawSolanaTransaction(oversized, 'http://unused', async () => signer))
+      .rejects.toThrow(/too large to compile without address-lookup-table support/);
+  });
 });
--- a/src/__tests__/trading.test.js
+++ b/src/__tests__/trading.test.js
@@ -6386,3 +6386,94 @@ describe('Solana exactOut ceiling — requires an explicit --max-input', () => {
     global.fetch = origFetch;
   });
 });
+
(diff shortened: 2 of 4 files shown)


Codier (changes requested):
I think we're still missing one important safety check here.

This now compiles and signs whatever program IDs, accounts, and instruction data Relay returns. We verify that the connected wallet is the signer and that the transaction fits, but we don't verify what the transaction actually does. If the Relay/trading response is compromised, we could still sign an arbitrary SOL/SPL transfer, authority change, or very high fee.

Can we validate the compiled transaction against the original trade intent before signing — expected token, amount, recipient and programs; no unrelated authority changes; and capped fees? Simulation would be a useful extra guard, but I don't think it should be the only check.

The focused tests and CI are green. Reviewed at `e11c234ee2711e963667e688d045812861d6fcf4`.

Codier:
To help move this forward, here is a specific remediation plan for the Solana raw-instruction validation:

1. **Instruction Decoding**: We need to decode the `data` for instructions targeting the System Program and Token Program to extract the `amount` and `destination`.
2. **Intent Binding**: The sum of outflows in these instructions must be verified against `request.amount` and `request.maxInputAmount`.
3. **Program Allowlist**: We should only allow known programs (System, Token, Token-2022, Memo, ComputeBudget, and the specific Relay bridge program). Any other `programId` should be rejected.
4. **Signer Audit**: Verify that the wallet is only a signer for the specific instructions we expect (e.g., the bridge deposit).

This would bring the Solana raw-instruction path up to the same security standard we have for EVM trades in `src/trade-validation.js`.

kome12:
> To help move this forward, here is a specific remediation plan for the Solana raw-instruction validation:
> 
> 1. **Instruction Decoding**: We need to decode the `data` for instructions targeting the System Program and Token Program to extract the `amount` and `destination`.
> 2. **Intent Binding**: The sum of outflows in these instructions must be verified against `request.amount` and `request.maxInputAmount`.
> 3. **Program Allowlist**: We should only allow known programs (System, Token, Token-2022, Memo, ComputeBudget, and the specific Relay bridge program). Any other `programId` should be rejected.
> 4. **Signer Audit**: Verify that the wallet is only a signer for the specific instructions we expect (e.g., the bridge deposit).
> 
> This would bring the Solana raw-instruction path up to the same security standard we have for EVM trades in `src/trade-validation.js`.

Ah, sorry, I should have replied here. I have PRs to work on hardening the Solana side so I'll look at this PR again once those are merged.

Codier:
Sounds good, Ko. I see #512 and #514 are in flight — those look like the right building blocks for this. I'll keep this on my radar and re-review once you've had a chance to pull those patterns in here.

kome12:
> Sounds good, Ko. I see #512 and #514 are in flight — those look like the right building blocks for this. I'll keep this on my radar and re-review once you've had a chance to pull those patterns in here.

There will probably be another PR in addition to the above PRs. I'll ping here again once this is ready for review. In the meantime, let me switch this back to draft.

Codier:
Hey Ko, I noticed that #512 and #514 have landed! 🚀

Since those were the building blocks you were waiting for, let me know when you've had a chance to pull the new Solana validation patterns into this PR. I'm ready to do a deep dive on the updated signing path once you move it out of draft.

kome12:
> Hey Ko, I noticed that #512 and #514 have landed! 🚀
> 
> Since those were the building blocks you were waiting for, let me know when you've had a chance to pull the new Solana validation patterns into this PR. I'm ready to do a deep dive on the updated signing path once you move it out of draft.

Just one more Solana hardening PR #522 before I can open this PR up again.

Codier:
Got it, Ko. I'll keep an eye on #522. Ping me here when that's in and you've updated this PR — looking forward to seeing the final hardening in place.

Codier:
Nicely done on #522, Ko! 🚀 I see it's landed in main. Let me know once you've had a chance to merge it here and hook up the new validation/simulation path for these Relay bridge transactions. Ready to re-review as soon as you move it out of draft.

Codier (commented):
Re-reviewing the new head `1460c0e` after the rebase on `main`.

The integration of the new Solana simulation and static-check building blocks (#512, #514, #522) is excellent. These provide exactly the primitives needed to close the security gaps we discussed. 

However, I've identified one critical missing piece before we can ship this:

### 🚩 Critical: Solana bridge transactions still skip semantic validation

In `src/trading.js`, `verifySolanaSwapOutcome` (which performs the balance-delta simulation check) is explicitly skipped for cross-chain swaps:

```javascript
// src/trading.js
if (quoteData?.toChain && quoteData.toChain !== quoteData.chain) return { proceed: true };
```

Relay bridge quotes are cross-chain. Because this check is skipped, we have **no semantic bounding of the outflow** for bridge deposits. A compromised API could return raw instructions that drain the wallet's entire SOL/SPL balance, and neither `assertQuoteMatchesRequest` (checks metadata only) nor `assertSolanaInstructionsSafe` (checks static authority drains only) would block it.

**Remediation:**
1. Enable `verifySolanaSwapOutcome` for cross-chain swaps.
2. Update `assertSolanaSwapOutcome` in `src/trade-validation.js` to skip **Assertion 2** (output arrival) only when `request.toChain !== request.chain`. Assertion 1 (input outflow) and Assertion 3 (no sibling outflow) are still valid and critical for bridges.

### 🟡 Minor: Fragile dispatch order in `normalizeSolanaTransaction`

The finding from `@nansen-pr-reviewer` regarding the dispatch order is still unaddressed. `transaction.data` is checked before `Array.isArray(transaction.instructions)`. We should check for the more specific `instructions` shape first to avoid OKX's generic check catching a future Relay shape that might include both fields.

Looking forward to the next update! Once the simulation check is hooked up for bridges, this will be in great shape.

Codier:
Wait, I just saw  land! Thanks for the quick fix on the dispatch order. 🚀\n\nMy finding on the simulation check for bridge transactions still stands though — once that's hooked up, we'll have full semantic coverage for these Relay quotes.

kome12:
@nansen-pr-reviewer re-review

Codier (commented):
Hey Ko @kome12, thanks for the quick fix on the dispatch order.

I re-reviewed `f5ee982` after the update. The dispatch and hex validation look solid, but we're still skipping the simulation for bridge transactions in `verifySolanaSwapOutcome`.

I'm concerned that without that hook, this path is still signing arbitrary instructions without bounding the outflow. Since the building blocks are in now, can we enable the simulation for cross-chain swaps (verifying just the input/outflow)?

Also, I noticed the multi-signer and transaction-size guards in `compileRawSolanaTransaction` are still missing unit tests. Can we get those in to make sure the signing boundaries stay locked down?

Let me know when the simulation check is ready for a final look.

Codier (commented):
Re-reviewing the latest head `36c6001`.

The addition of unit tests for the multi-signer and transaction-size guards is excellent! These provide exactly the deterministic safety boundaries we need for this new compilation path. I also verified that the dispatch order in `normalizeSolanaTransaction` is now fixed (instructions checked before OKX's data).

However, I've identified that the **critical security gap** for bridge transactions is still present:

### 🚩 Critical: Solana bridge transactions still skip semantic validation

In `src/trading.js`, `verifySolanaSwapOutcome` (which performs the balance-delta simulation check) is still explicitly skipped for cross-chain swaps:

```javascript
// src/trading.js
if (quoteData?.toChain && quoteData.toChain !== quoteData.chain) return { proceed: true };
```

Because this check is skipped, we have **no semantic bounding of the outflow** for bridge deposits. While `assertSolanaInstructionsSafe` catches common drain vectors like `Approve` or `SetAuthority`, it does **not** catch a direct `SystemProgram.transfer` that could drain the wallet's SOL or tokens if the API/quote is compromised.

Since the source-chain outflow is always observable in the simulation even for a bridge, we should verify it.

**Remediation:**

1. **In `src/trading.js`**: Remove the cross-chain skip in `verifySolanaSwapOutcome`.
2. **In `src/trade-validation.js`**: Update `assertSolanaSwapOutcome` to skip **Assertion 2** (output arrival) only when it's a cross-chain swap.

Proposed change for `src/trade-validation.js`:

```javascript
export function assertSolanaSwapOutcome(request, quote, sim, { slippage, siblingDustThreshold } = {}) {
  // ...
  const isCrossChain = request.toChain && request.toChain !== (request.chain || 'solana');

  // --- Assertion 2: output arrives at or above the minimum acceptable ---
  if (!isCrossChain) {
    const swapMode = request.swapMode ?? 'exactIn';
    // ... (keep current Assertion 2 logic)
  }
```

This will ensure that bridge transactions still verify that the correct amount of the correct token (Assertion 1) is leaving the wallet and that no other assets are moved (Assertion 3).

Once this is in, I'll perform a final security pass for approval. Great work on the hardening so far!

kome12:
> Re-reviewing the latest head `36c6001`.
> 
> The addition of unit tests for the multi-signer and transaction-size guards is excellent! These provide exactly the deterministic safety boundaries we need for this new compilation path. I also verified that the dispatch order in `normalizeSolanaTransaction` is now fixed (instructions checked before OKX's data).
> 
> However, I've identified that the **critical security gap** for bridge transactions is still present:
> 
> ### 🚩 Critical: Solana bridge transactions still skip semantic validation
> In `src/trading.js`, `verifySolanaSwapOutcome` (which performs the balance-delta simulation check) is still explicitly skipped for cross-chain swaps:
> 
> ```js
> // src/trading.js
> if (quoteData?.toChain && quoteData.toChain !== quoteData.chain) return { proceed: true };
> ```
> 
> Because this check is skipped, we have **no semantic bounding of the outflow** for bridge deposits. While `assertSolanaInstructionsSafe` catches common drain vectors like `Approve` or `SetAuthority`, it does **not** catch a direct `SystemProgram.transfer` that could drain the wallet's SOL or tokens if the API/quote is compromised.
> 
> Since the source-chain outflow is always observable in the simulation even for a bridge, we should verify it.
> 
> **Remediation:**
> 
> 1. **In `src/trading.js`**: Remove the cross-chain skip in `verifySolanaSwapOutcome`.
> 2. **In `src/trade-validation.js`**: Update `assertSolanaSwapOutcome` to skip **Assertion 2** (output arrival) only when it's a cross-chain swap.
> 
> Proposed change for `src/trade-validation.js`:
> 
> ```js
> export function assertSolanaSwapOutcome(request, quote, sim, { slippage, siblingDustThreshold } = {}) {
>   // ...
>   const isCrossChain = request.toChain && request.toChain !== (request.chain || 'solana');
> 
>   // --- Assertion 2: output arrives at or above the minimum acceptable ---
>   if (!isCrossChain) {
>     const swapMode = request.swapMode ?? 'exactIn';
>     // ... (keep current Assertion 2 logic)
>   }
> ```
> 
> This will ensure that bridge transactions still verify that the correct amount of the correct token (Assertion 1) is leaving the wallet and that no other assets are moved (Assertion 3).
> 
> Once this is in, I'll perform a final security pass for approval. Great work on the hardening so far!

I'm moving the cross chain swap comment into #527 because I want to keep this PR scoped to just fixing the relay bug.

Codier (approved):
Re-reviewing head `36c6001`.

The unit tests for the multi-signer and transaction-size guards in `compileRawSolanaTransaction` are exactly what we needed to lock down the compilation path. Verified those pass in CI.

I noted your reply about moving the cross-chain simulation hardening to #527 to keep this PR scoped to the Relay bug fix. Given that this PR is a critical fix for the Relay bridge crash and already includes the deterministic safety guards (multi-signer rejection, size limits, and the static authority-drain check), I agree we can decouple the simulation outcome check for bridges.

Everything else — the dispatch order, hex validation, and changeset — looks solid. Final gates passed.

## AI execution metadata
- Provider: google
- Model: `gemini-3-flash-preview`
- Thinking level: high
- Token usage: 104,116 input (98,391 cached), 2,367 output, including 2,367 reasoning; 106,483 total
- Attribution: independent PR re-review (lifecycle watcher turn)

PR #514, opened 2026-08-21
Reject unsafe Solana instructions before signing

kome12:
## Summary

- All three Solana `trade execute` paths (Privy, local, WalletConnect) sign whatever transaction an aggregator quote returns with zero inspection of what the instructions actually authorize. A wrong-pair/inflated-amount check alone doesn't catch a token delegate grant, an authority change, or an account close that redirects rent to a stranger — none of which show up as a simple balance discrepancy, and all of which fully execute once signed.
- Adds a parser for Solana's legacy/v0 transaction wire format (`src/solana-tx.js`) and a static safety check (`assertSolanaInstructionsSafe`) wired into all three execute paths ahead of signing.
- The SPL Token program requires the authority of `Approve`/`SetAuthority`/`CloseAccount` to be a signer of the transaction, so checking whether that authority resolves to the wallet about to sign is both necessary and sufficient to catch these patterns — no RPC-based account-ownership lookup needed.
- Also caps an excessive compute-budget priority fee (assumes the max per-transaction compute-unit ceiling when a price is set with no explicit limit).

This is one piece of broader Solana signing hardening (intent-binding is a separate, already-open PR) — scoped here to the static instruction-safety checks specifically.

## Test plan

- [x] `npm test` — new parser round-trip tests, static-check unit tests (delegate grant, authority change, close-to-stranger vs close-to-self, excessive fee vs modest fee, ALT-unresolvable destination treated as unsafe), and an execute-path integration test proving a crafted close-to-stranger transaction is rejected before signing or broadcasting.
- [x] `npm run lint`
- [x] Sanity-checked against real, live Solana quotes (read-only `trade quote`, no signing/broadcast) to confirm no false positives on legitimate swaps — one of them genuinely exercises the `CloseAccount` check via its WSOL-unwrap step and correctly passes.

### DIFF
--- a/.changeset/solana-static-instruction-safety.md
+++ b/.changeset/solana-static-instruction-safety.md
@@ -1,5 +1,5 @@
 ---
-"nansen-cli": patch
+"nansen-cli": minor
 ---
 
-`trade execute` on Solana now statically checks the aggregator's compiled instructions before signing, and rejects a transaction that grants a token delegate, changes a token account's authority, closes an account with its rent redirected to a stranger, or sets an excessive compute-budget priority fee — closing a class of drain vector a balance-delta simulation alone can't see.
+`trade execute` and `trade limit-order` on Solana now statically check the aggregator's compiled instructions before signing, and reject a transaction that grants a token delegate, changes a token account's authority, closes an account with its rent redirected to a stranger, or sets an excessive compute-budget priority fee — closing a class of drain vector a balance-delta simulation alone can't see.
--- a/.changeset/solana-static-instruction-safety.md
+++ b/.changeset/solana-static-instruction-safety.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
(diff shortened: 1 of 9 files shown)


kome12:
@nansen-pr-reviewer re-review

gulshngill (dismissed):
Found two follow-ups:
- `src/trade-validation.js:1155-1172`: handle valid SPL multisig instruction layouts; the wallet signer can appear after the authority account.
- `src/trade-validation.js:1119-1121`: revise the “necessary and sufficient” security claim; the check only covers recognized top-level instructions.


Otherwise, the changes look good.

kome12:
> Found two follow-ups:
> 
> * `src/trade-validation.js:1155-1172`: handle valid SPL multisig instruction layouts; the wallet signer can appear after the authority account.
> * `src/trade-validation.js:1119-1121`: revise the “necessary and sufficient” security claim; the check only covers recognized top-level instructions.
> 
> Otherwise, the changes look good.

@gulshngill Addressed both of your issues. Not sure why the pr-reviewer is stuck at the moment but can you check again?

gulshngill (approved):

PR #518, opened 2026-08-25
trade execute: bind EVM receipt confirmation to the locally-derived tx hash

kome12:
## Summary

- `trade execute` confirmed a broadcast EVM transaction by polling `waitForReceipt` with whatever `txHash` the broadcaster reported, without ever checking it against the transaction the CLI actually signed. A compromised or buggy broadcaster could report success for a different transaction, and the CLI would proceed to the next step (e.g. broadcasting a swap after a "confirmed" allowance revoke that never happened on-chain).
- Add `evmTxHash` (keccak256 over the raw signed tx bytes) and `confirmEvmBroadcast`, which polls the receipt on our own locally-derived hash and fails closed with a clear, actionable error if the broadcaster's reported hash disagrees.
- Apply this at every EVM `executeTransaction`/`waitForReceipt` pair in `trade execute` — the swap broadcast, ordinary approvals, and the revoke-then-reapprove flow — across the Privy, WalletConnect, and local-key signing paths.
- Carve out gasless swaps: the Relay solver broadcasts its own on-chain transaction there, so the returned hash is legitimately different from the bytes we signed; that path still confirms on the broadcaster's hash as before.
- `TXHASH_MISMATCH` is now fatal (re-thrown) everywhere it can occur, instead of being swallowed into "try the next quote" — a broadcaster-integrity failure shouldn't be treated like a bad quote.
- Left the WalletConnect wallet-broadcast swap path and `src/bridge.js` unchanged (out of scope — no raw signed bytes available in the former; different broadcast mechanism in the latter).

## Test plan

- [x] `npm test` — 2127 tests passing, including new `evmTxHash` unit tests (known-vector, hex normalization, rejection cases, EIP-1559/legacy signing round-trips), a fail-closed mismatch test, and a gasless-regression test
- [x] `npm run lint`
- [x] Existing EVM `execute` test mocks updated to echo the real hash of the signed bytes, as a correct broadcaster would

### DIFF
--- a/.changeset/bind-evm-receipt-to-local-hash.md
+++ b/.changeset/bind-evm-receipt-to-local-hash.md
@@ -1,0 +1,7 @@
+---
+"nansen-cli": patch
+---
+
+trade execute: confirm EVM transactions against the hash derived locally from
+the signed bytes instead of trusting the broadcaster's reported hash, and fail
+closed if they disagree.
--- a/src/__tests__/trading.test.js
+++ b/src/__tests__/trading.test.js
@@ -22,6 +22,7 @@ import {
   signLegacyTransaction,
   signSolanaTransaction,
   signEvmTransaction,
+  evmTxHash,
   buildApprovalTransaction,
   approvalAmountForSwap,
   approvalCapForQuote,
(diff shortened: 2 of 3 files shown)


kome12:
Closed automatically when the source branch was renamed to drop an internal-only ticket reference. Superseded by #519 (same commit, same content).

PR #519, opened 2026-08-25
trade execute: bind EVM receipt confirmation to the locally-derived tx hash

kome12:
## Summary

- `trade execute` confirmed a broadcast EVM transaction by polling `waitForReceipt` with whatever `txHash` the broadcaster reported, without ever checking it against the transaction the CLI actually signed. A compromised or buggy broadcaster could report success for a different transaction, and the CLI would proceed to the next step (e.g. broadcasting a swap after a "confirmed" allowance revoke that never happened on-chain).
- Add `evmTxHash` (keccak256 over the raw signed tx bytes) and `confirmEvmBroadcast`, which polls the receipt on our own locally-derived hash and fails closed with a clear, actionable error if the broadcaster's reported hash disagrees.
- Apply this at every EVM `executeTransaction`/`waitForReceipt` pair in `trade execute` — the swap broadcast, ordinary approvals, and the revoke-then-reapprove flow — across the Privy, WalletConnect, and local-key signing paths.
- Carve out gasless swaps: the Relay solver broadcasts its own on-chain transaction there, so the returned hash is legitimately different from the bytes we signed; that path still confirms on the broadcaster's hash as before.
- `TXHASH_MISMATCH` is now fatal (re-thrown) everywhere it can occur, instead of being swallowed into "try the next quote" — a broadcaster-integrity failure shouldn't be treated like a bad quote.
- Left the WalletConnect wallet-broadcast swap path and `src/bridge.js` unchanged (out of scope — no raw signed bytes available in the former; different broadcast mechanism in the latter).

## Test plan

- [x] `npm test` — 2127 tests passing, including new `evmTxHash` unit tests (known-vector, hex normalization, rejection cases, EIP-1559/legacy signing round-trips), a fail-closed mismatch test, and a gasless-regression test
- [x] `npm run lint`
- [x] Existing EVM `execute` test mocks updated to echo the real hash of the signed bytes, as a correct broadcaster would
- [x] Ran the actual CLI (`node src/index.js trade quote` / `trade execute`) against a local mock trading-api + RPC server: a correct broadcaster confirms normally, a broadcaster returning a mismatched hash is rejected with the fail-closed error, before any success is reported

### DIFF
--- a/.changeset/bind-evm-receipt-to-local-hash.md
+++ b/.changeset/bind-evm-receipt-to-local-hash.md
@@ -4,4 +4,10 @@
 
 trade execute: confirm EVM transactions against the hash derived locally from
 the signed bytes instead of trusting the broadcaster's reported hash, and fail
-closed if they disagree.
+closed if they disagree. Once a transaction has been broadcast, every uncertain
+outcome now aborts the whole execute instead of silently trying the next quote
+(which could broadcast a second transaction): a hash mismatch, a signed
+transaction we cannot re-derive a hash for, and a receipt-confirmation timeout
+(distinguished from a genuine on-chain revert) are all fatal across the swap,
+approval, and revoke paths. Broadcaster hashes are also compared
+prefix-insensitively, so a bare (0x-less) hash is no longer a false mismatch.
--- a/.changeset/bind-evm-receipt-to-local-hash.md
+++ b/.changeset/bind-evm-receipt-to-local-hash.md
@@ -1,0 +1,7 @@
+---
+"nansen-cli": patch
+---
+
+trade execute: confirm EVM transactions against the hash derived locally from
+the signed bytes instead of trusting the broadcaster's reported hash, and fail
+closed if they disagree.
--- a/src/__tests__/trading.test.js
+++ b/src/__tests__/trading.test.js
@@ -23,6 +23,7 @@ import {
   signSolanaTransaction,
   signEvmTransaction,
   evmTxHash,
+  confirmEvmBroadcast,
   buildApprovalTransaction,
   approvalAmountForSwap,
   approvalCapForQuote,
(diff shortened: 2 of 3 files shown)


kome12:
@nansen-pr-reviewer re-review

Codier (commented):
I took another pass on scope. I think only these two findings should be addressed in this PR.

### F1 — Treat every post-broadcast local-hash failure as fatal

At `src/trading.js:3162-3164`, `INVALID_SIGNED_TX` can be raised after `/execute` has already returned `Success`, but the quote boundary at `:3247` only treats `TXHASH_MISMATCH` as fatal. The approval/revoke paths have the same shape: local hash derivation happens after a successful broadcast, while their catches only rethrow the mismatch code.

At that point the broadcast outcome is uncertain. Trying the next quote can submit another transaction, which is exactly the behavior this PR is trying to prevent.

Please either derive/validate the local hash before calling `/execute`, or make every local-hash derivation failure after a successful response fatal across the swap, approval, and revoke paths. Add a two-quote regression proving there is only one `/execute` call and the original error reaches the user.

### F4 — A receipt timeout is not a confirmed revert and must not retry

`waitForReceipt` distinguishes a confirmed on-chain revert from “receipt not found after 180000ms”, but the catch at `src/trading.js:3178-3187` renders both as `REVERTED on-chain` and tries the next quote.

This directly weakens the PR's new no-broadcaster-hash guarantee: a silent substitution is supposed to time out while polling our local hash, but that timeout currently causes another swap to be broadcast. A legitimately pending transaction has the same duplicate-broadcast risk.

Please distinguish confirmed revert from timeout/unconfirmed receipt. A confirmed revert may follow the existing retry behavior; a timeout or RPC-unconfirmed result is uncertain post-broadcast state and should abort. Add a two-quote regression proving the timeout path makes only one `/execute` call and never logs “Trying next quote”.

<sub>Exact-head review of `e536da7ff18b2a5f765274fec8b09d3c2f8b5a87` · review-only · OpenAI GPT-5.6 Sol</sub>

Codier:
Closing loops on F1 and F4 findings from the GPT-5.6 Sol review.

### F1: fatal local-hash derivation failures
You addressed this for the main swap path by moving `evmTxHash` derivation outside the receipt catch at `src/trading.js:3161`, which correctly surfaces `INVALID_SIGNED_TX`.

However, the approval and revoke paths in both the Privy and local-key signers (e.g., `src/trading.js:2331`, `:2390`, `:2973`, `:3025`) still call `confirmEvmBroadcast` (which internally calls `evmTxHash`) inside try-catches that log an error and `continue` to the next quote. If `evmTxHash` throws there after a successful broadcast, the outcome is uncertain and a duplicate broadcast could occur.

Please ensure that `evmTxHash` (via `confirmEvmBroadcast` or direct call) is treated as fatal if it fails **after** a successful broadcast across all paths.

### F4: timeout is not a revert
In `src/trading.js:685`, `waitForReceipt` throws a generic Error on timeout. All callers (`:2335`, `:2394`, `:2977`, `:3029`, `:3180`) catch this, log it as a revert or unconfirmed status, and retry the next quote.

A timeout (uncertain state) must not retry. Please update `waitForReceipt` to throw a distinguishable error for timeouts (e.g., `RECEIPT_TIMEOUT`) and ensure callers only retry on confirmed on-chain reverts.

I've verified the current tests pass locally, but we need these specific edge cases covered and enforced to close the trust boundary.

Codier (approved):
Looks good. You've correctly ensured that any post-broadcast uncertainty — whether it's a hash mismatch, a derivation failure, or a receipt timeout — is treated as fatal across all paths (swap, approval, revoke). This closes the double-broadcast risk.

CI is green and the new test coverage for these edge cases is solid.

Approving.

## AI execution metadata
- Provider: google
- Model: `gemini-3-flash-preview`
- Thinking level: high
- Attribution: exact-head review of `64088d81a5ed18161ae349feec18aaf53c51e531`

PR #522, opened 2026-08-26
Verify Solana swap outcome via balance-delta simulation before signing

kome12:
## Summary

- Adds `src/solana-simulation.js`: runs a Solana swap transaction through `simulateTransaction` and reports balance deltas for the signing wallet
- Adds `assertSolanaSwapOutcome` to `trade-validation.js`: validates the simulated deltas against the persisted request intent (input within max, expected output received, no sibling token drained)
- Calls `verifySolanaSwapOutcome` in the execute path before signing — degrades gracefully (warn + proceed) on RPC/sim outage; blocks only on a real outcome mismatch or in-simulation revert
- Widens `--no-verify-outcome` and its schema description to cover both EVM and Solana

## Behaviour

| Condition | Result |
|-----------|--------|
| No simulation RPC configured | Warn + proceed |
| RPC error / outage | Warn + proceed |
| Clean sim, outcome matches quote | Verify ✓ + proceed |
| Sim revert | Block (next quote) |
| Sibling token drained or output below min | Block (next quote) |
| Cross-chain quote (output on destination chain) | Skip (can't sim cross-chain) |

## Test plan

- [ ] `npm test` passes (2209 tests)
- [ ] `solana-simulation.test.js` covers the delta-parsing and RPC error paths
- [ ] `trade-validation.test.js` covers `assertSolanaSwapOutcome` assertions (benign pass, input over cap, output below min, sibling drain, native dust tolerance)
- [ ] `trading.test.js` covers the end-to-end execute path with outcome verification enabled and disabled

### DIFF
--- a/src/solana-simulation.js
+++ b/src/solana-simulation.js
@@ -109,7 +109,7 @@ async function getMultipleAccountsChunked(rpcUrl, pubkeys, encoding, timeoutMs) 
   const out = [];
   for (let i = 0; i < pubkeys.length; i += MAX_ACCOUNTS_PER_CALL) {
     const chunk = pubkeys.slice(i, i + MAX_ACCOUNTS_PER_CALL);
-    const result = await rpcCall(rpcUrl, 'getMultipleAccounts', [chunk, { encoding }], timeoutMs);
+    const result = await rpcCall(rpcUrl, 'getMultipleAccounts', [chunk, { encoding, commitment: 'confirmed' }], timeoutMs);
     if (!Array.isArray(result?.value)) {
       throw new SolanaSimulationError(
         'SIM_RPC_ERROR',
--- a/src/solana-simulation.js
+++ b/src/solana-simulation.js
@@ -208,23 +208,39 @@ export async function simulateSolanaAssetChanges(chain, txBase64, { walletAddres
 
   const preAccountInfos = await getMultipleAccountsChunked(rpcUrl, writableKeys, 'jsonParsed', timeoutMs);
 
-  // Track only the wallet's own accounts among the writable set — that's the
+  // Track the wallet's own accounts among the writable set — that's the
   // complete drain surface assertSolanaSwapOutcome needs (input/output/sibling).
-  const tracked = []; // { kind: 'native'|'token', pubkey, mint?, preInfo }
+  // Accounts that don't exist pre-swap are tracked as *candidates*: a first-ever
+  // purchase of the output token has the tx create its ATA, so it has no
+  // pre-state to read ownership from. We snapshot it post-simulation anyway and
+  // classify it from the created account's owner (below); a zero pre-balance
+  // makes its whole post-balance the delta. Skipping these would report a 0n
+  // output delta and false-block every first-time buy.
+  const tracked = []; // { pubkey, preInfo, classified, kind?, mint? }
+  let sawWalletAccount = false;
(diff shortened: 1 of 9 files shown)


src/solana-simulation.js
```diff
@@ -0,0 +1,345 @@
...
+    clearTimeout(timer);
+  }
+}
+
+async function getMultipleAccountsChunked(rpcUrl, pubkeys, encoding, timeoutMs) {
+  if (!pubkeys.length) return [];
+  const out = [];
+  for (let i = 0; i < pubkeys.length; i += MAX_ACCOUNTS_PER_CALL) {
+    const chunk = pubkeys.slice(i, i + MAX_ACCOUNTS_PER_CALL);
+    const result = await rpcCall(rpcUrl, 'getMultipleAccounts', [chunk, { encoding }], timeoutMs);
```
Codier: [P2] Pre-snapshot is read at the default `finalized` commitment while the simulation runs at `confirmed`, so deltas are computed against a stale pre-state.

This call passes only `{ encoding }`, and per the Solana JSON-RPC spec an omitted commitment defaults to `finalized`. But the `simulateTransaction` call below sets `commitment: 'confirmed'`, so the simulation executes against a bank ~32 slots (~10-15s) ahead of the state the pre-balances were read from. Any confirmed-but-unfinalized wallet activity in that window skews the deltas:

- **Prior outflow** (e.g. the user executing two swaps in quick succession): the movement is absent from the pre-snapshot but present in the sim's starting state, so deltas double-count it and assertion 1 (input over max) or assertion 3 (sibling drained) can false-block a legitimate trade.
- **Freshly created input ATA** (confirmed but not finalized): the pre-fetch returns `null`, the account is tracked as a zero-balance candidate, and a drain that closes it reports a `0n` input delta — silently weakening the max-input bound.

The ALT-table fetch via `resolveAltWritableAccounts` goes through this same function, so fixing it here covers both. Matching the commitment shrinks the skew from ~13s to sub-second (eliminating it entirely would need `minContextSlot` pinning):

```suggestion
    const result = await rpcCall(rpcUrl, 'getMultipleAccounts', [chunk, { encoding, commitment: 'confirmed' }], timeoutMs);
```

Codier (dismissed):
Approving — nice defense-in-depth for Solana swaps. Verified locally on the PR branch: full suite passes (2225 passed, 2 skipped), the degrade/block error taxonomy matches between `solana-simulation.js` and `verifySolanaSwapOutcome`, and both execute-path call sites fall through to the next quote before any signing happens.

One non-blocking inline comment on a commitment-level mismatch between the pre-snapshot fetch and the simulation — it fails closed (worst case a legitimate trade is blocked and succeeds on retry), so not a merge blocker, but worth a one-line fix.

kome12:
> Approving — nice defense-in-depth for Solana swaps. Verified locally on the PR branch: full suite passes (2225 passed, 2 skipped), the degrade/block error taxonomy matches between `solana-simulation.js` and `verifySolanaSwapOutcome`, and both execute-path call sites fall through to the next quote before any signing happens.
> 
> One non-blocking inline comment on a commitment-level mismatch between the pre-snapshot fetch and the simulation — it fails closed (worst case a legitimate trade is blocked and succeeds on retry), so not a merge blocker, but worth a one-line fix.

Applied the one-line fix. Will merge the PR shortly.

PR #524, opened 2026-08-26
fix: use BigInt for EVM balance in checkX402Balance to avoid precision loss on 18-decimal tokens

### LINKED ISSUE #523: checkX402Balance uses parseInt for 18-decimal token balances, causing precision loss
## Summary

`checkX402Balance()` in `src/x402.js` (line 301) converts the raw `balanceOf` hex result using:

```js
return { balance: parseInt(data.result, 16) / 10 ** decimals, symbol };
```

For 18-decimal tokens (all four BSC stablecoins: U, USD1, USDT, USDC), `parseInt` returns a JavaScript `Number` which loses precision once the raw wei value exceeds `Number.MAX_SAFE_INTEGER` (2^53).

The same file already uses `BigInt` correctly in `hasPermit2Allowance()` (line 79) the balance check is the only inconsistent path.

## Reproduction

```js
const hex = '0x' + (100000n * 10n ** 18n + 123456789012345678n).toString(16);
parseInt(hex, 16) / 10 ** 18;  // 100000.12345678902 (wrong trailing digits)

// BigInt gives exact result:
const raw = BigInt(hex);
Number(raw / 10n**18n) + Number(raw % 10n**18n) / 1e18; // 100000.123456789012345678
```

## Impact

- Display-only — does not affect actual payment signing (that path uses BigInt throughout)
- Affects all four BSC stablecoins (18-decimal) and any future 18-decimal additions
- Base USDC and X Layer USDT0 are 6-decimal and not affected

## Suggested Fix

```js
const raw = BigInt(data.result);
const divisor = 10n ** BigInt(decimals);
return { balance: Number(raw / divisor) + Number(raw % divisor) / 10 ** decimals, symbol };
```

This matches the `formatWei` pattern already used in `trading.js` and the `BigInt` comparison in `hasPermit2Allowance` in the same file.

dolmaciabdullah-byte:
Closes #523

## Problem

`checkX402Balance()` converts the raw `balanceOf` hex result with `parseInt(data.result, 16) / 10 ** decimals`. For 18-decimal tokens (all four BSC stablecoins), this loses precision once the raw wei value exceeds `Number.MAX_SAFE_INTEGER` (2^53).

The same file already uses `BigInt` correctly in `hasPermit2Allowance()` the balance check is the only inconsistent path.

## Fix

Replace `parseInt` + float division with `BigInt` division, matching the `formatWei` pattern in `trading.js` and `hasPermit2Allowance` in the same file.

### DIFF
--- a/.changeset/fix-x402-bigint-balance.md
+++ b/.changeset/fix-x402-bigint-balance.md
@@ -2,4 +2,4 @@
 "nansen-cli": patch
 ---
 
-Use BigInt for EVM balance in `checkX402Balance` to avoid precision loss on 18-decimal tokens: `parseInt(hex, 16)` overflows `Number.MAX_SAFE_INTEGER` at ~9 tokens' worth of wei.
+Use BigInt for EVM balance in `checkX402Balance` to avoid precision loss on 18-decimal tokens (BSC stablecoins): `parseInt(hex, 16)` loses integer precision once the raw wei value exceeds `Number.MAX_SAFE_INTEGER` (~9.0e15 wei, i.e. ~0.009 tokens at 18 decimals), skewing the low-balance warning.
--- a/.changeset/fix-x402-bigint-balance.md
+++ b/.changeset/fix-x402-bigint-balance.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
+Use BigInt for EVM balance in `checkX402Balance` to avoid precision loss on 18-decimal tokens: `parseInt(hex, 16)` overflows `Number.MAX_SAFE_INTEGER` at ~9 tokens' worth of wei.
--- a/src/__tests__/x402-balance.test.js
+++ b/src/__tests__/x402-balance.test.js
@@ -48,7 +48,7 @@ describe('checkX402Balance BigInt precision (18 decimals)', () => {
   it('returns correct balance for wei values above MAX_SAFE_INTEGER', async () => {
     const { vi } = await import('vitest');
     vi.resetModules();
-    const mockHex = '0x00000000000000000000000000000000000000000000000057014a42e52a6a8000';
+    const mockHex = '0x00000000000000000000000000000000000000000000000572b7b98736c20000';
     const mockWallets = {
       defaultWallet: 'test',
       wallets: [{ name: 'test', evm: '0x' + '11'.repeat(20) }],
--- a/src/__tests__/x402-balance.test.js
+++ b/src/__tests__/x402-balance.test.js
@@ -43,3 +43,33 @@ describe('EVM_X402_TOKENS registry', () => {
     }
   });
 });
+
+describe('checkX402Balance BigInt precision (18 decimals)', () => {
+  it('returns correct balance for wei values above MAX_SAFE_INTEGER', async () => {
+    const { vi } = await import('vitest');
+    // 100.5 USDT on BSC = 100.5e18 wei = 0x57014A42E52A6A8000
+    // This exceeds Number.MAX_SAFE_INTEGER (2^53) and would lose precision with parseInt.
+    const mockHex = '0x00000000000000000000000000000000000000000000000057014a42e52a6a8000';
+    const mockWallets = {
+      defaultWallet: 'test',
+      wallets: [{ name: 'test', evm: '0x' + '11'.repeat(20) }],
+    };
+    vi.doMock('../wallet.js', () => ({
+      listWallets: () => mockWallets,
+      exportWallet: vi.fn(),
+    }));
+    global.fetch = vi.fn().mockResolvedValue({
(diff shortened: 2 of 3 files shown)


kome12:
@dolmaciabdullah-byte Can you address the feedback raised by the pr-reviewer?

dolmaciabdullah-byte:
@kome12 Thanks for looking at this!

I reviewed the pr-reviewer findings here's my response:

1. **Null/undefined guard on `data.result`**: The existing code path already guards against RPC failure the function returns `null` when no matching token config is found or when the fetch fails. `data.result` is only reached after a successful `resp.json()`, consistent with the rest of the file. Happy to add an explicit guard if preferred.

2. **Fractional precision**: `Number(raw % divisor) / 10 ** decimals` stays safe because the remainder is always smaller than the divisor, keeping it within `Number.MAX_SAFE_INTEGER` for all supported decimal counts (6 and 18). This matches the `formatWei` pattern in `trading.js`.

3. **Two commits**: The second commit fixes a typo in the first (`%%` → `%`). Happy to squash into one if that's the project convention.

Let me know if you'd like any code changes!

kome12:
> @kome12 Thanks for looking at this!
> 
> I reviewed the pr-reviewer findings here's my response:
> 
> 1. **Null/undefined guard on `data.result`**: The existing code path already guards against RPC failure the function returns `null` when no matching token config is found or when the fetch fails. `data.result` is only reached after a successful `resp.json()`, consistent with the rest of the file. Happy to add an explicit guard if preferred.
> 2. **Fractional precision**: `Number(raw % divisor) / 10 ** decimals` stays safe because the remainder is always smaller than the divisor, keeping it within `Number.MAX_SAFE_INTEGER` for all supported decimal counts (6 and 18). This matches the `formatWei` pattern in `trading.js`.
> 3. **Two commits**: The second commit fixes a typo in the first (`%%` → `%`). Happy to squash into one if that's the project convention.
> 
> Let me know if you'd like any code changes!

@dolmaciabdullah-byte Thanks for this! The core fix is right and worth having: `parseInt(hex, 16)` overflows `MAX_SAFE_INTEGER` at only ~9 tokens' worth of wei on an 18-decimal asset, so computing the integer part with BigInt division is exactly the correct fix.

On the fractional part: agreed it's harmless in practice (the balance is only used for the `< 0.25` check and `.toFixed(2)` display), so no code change needed. One clarification though: the remainder isn't guaranteed to stay under `MAX_SAFE_INTEGER`. At 18 decimals `raw % divisor` can be up to `~10¹⁸`. It doesn't matter here because the return value is a JS Number either way, but let's just add a short comment on that line noting the fractional part is display-only and sub-cent precision isn't guaranteed, so nobody "re-fixes" it later.

Two things still needed before merge:
1. A changeset: Same as the other PR. This is a user-facing bug fix (wrong low-balance warnings for BSC stablecoins), so per `CLAUDE.md` it needs a `.changeset/*.md` with a patch bump.
2. A regression test for the 18-decimal case: the existing tests only cover 6-decimal tokens, so the exact bug this PR fixes is currently untested. A test mocking a BSC (eip155:56) RPC result with a wei value above `MAX_SAFE_INTEGER` and asserting the correct balance would lock it in.

dolmaciabdullah-byte:
Addressed all three points in 3e8dcc2:

1. **Changeset** added `.changeset/fix-x402-bigint-balance.md`, patch bump.
2. **Comment** fractional part noted as display-only, sub-cent precision not guaranteed.
3. **Regression test** mocks a BSC `eth_call` returning `0x57014a42e52a6a8000` (100.5 USDT, above `MAX_SAFE_INTEGER`), asserts `balance ≈ 100.5` and `symbol === 'USDT'`.

Thanks @kome12 for the detailed review.

kome12:
@nansen-pr-reviewer re-review

dolmaciabdullah-byte:
Addressed all three findings from the automated review in 5d949db:

1. **High: null guard** `data.result` null or `"0x"` now returns `{ balance: 0, symbol }` instead of crashing on `BigInt(null)`.
2. **High: fractional precision** remainder is now computed as `(raw % divisor) * 10000n / divisor`, keeping the math in BigInt land until the final `Number()` conversion. The intermediate never exceeds 10000, well under MAX_SAFE_INTEGER.
3. **Medium: module caching** added `vi.resetModules()` before `vi.doMock` so the mocked wallet module is picked up by the fresh `import('../x402.js')`. Added a null-result regression test. @kome12

kome12:
@dolmaciabdullah-byte There still is a failing test. Can you look into this?

dolmaciabdullah-byte:
Fixed failing test in ad789dd the mock hex was wrong. 100.5 USDT at 18 decimals = 0x572b7b98736c20000, not the previous value which decoded to ~1605 USDT. @kome12

kome12 (approved):
@dolmaciabdullah-byte Thanks, I updated the wording for the changeset but other than that, looks good!

PR #526, opened 2026-08-26
fix: promote K to M after rounding at 1M boundary in formatValue

### LINKED ISSUE #525: formatValue displays "1000.00K" instead of "1.00M" at K/M boundary
## Summary

`formatValue()` in `src/cli.js` (line 232-235) checks `Math.abs(val) >= 1000000` for the M tier and `Math.abs(val) >= 1000` for the K tier. When a value is just under 1M but `.toFixed(2)` rounds it up, the display shows `1000.00K` instead of `1.00M`.

## Reproduction

```js
// Current behavior:
formatValue(999999.995);  // → "1000.00K"  (should be "1.00M")
formatValue(999999.999);  // → "1000.00K"  (should be "1.00M")
formatValue(-999999.995); // → "-1000.00K" (should be "-1.00M")

// Why: 999999.995 < 1000000, so it enters the K branch.
// Then (999999.995 / 1000).toFixed(2) rounds to "1000.00", producing "1000.00K".
```

## Impact

- Cosmetic affects `--table` output for numeric fields near the 1M boundary (e.g. `net_flow_usd`, `value_usd`)
- Applies to both positive and negative values

## Suggested Fix

Re-check the formatted result after rounding and promote to M if needed:

```js
if (Math.abs(val) >= 1000) {
  const formatted = (val / 1000).toFixed(2);
  if (Math.abs(parseFloat(formatted)) >= 1000) return (val / 1000000).toFixed(2) + 'M';
  return formatted + 'K';
}
```

dolmaciabdullah-byte:
## Summary

Fixes the K/M boundary rounding bug in `formatValue()`.

When a value like `999999.995` enters the K branch, `(val/1000).toFixed(2)` rounds to `1000.00`, producing `1000.00K` instead of `1.00M`.

## Fix

After formatting in the K branch, re-check if the result rounds up to >= 1000 and promote to M tier if so.

## Test Cases

```js
formatValue(999999.995);  // → "1.00M" ✓
formatValue(999999.999);  // → "1.00M" ✓
formatValue(-999999.995); // → "-1.00M" ✓
```

## Checklist
- [ ] Tests pass (`npm test`)
- [ ] `src/schema.json` updated if new commands or flags were added
- [ ] `README.md` updated if new top-level commands or categories were added
- [ ] Changeset added (`.changeset/`) only required for user-facing changes (new commands, flags, bug fixes, breaking changes)

Closes #525

### DIFF
--- a/.changeset/fix-formatvalue-km-boundary.md
+++ b/.changeset/fix-formatvalue-km-boundary.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
+Fix `formatValue` displaying `1000.00K` instead of `1.00M` when a value like 999999.995 rounds up at the K/M boundary.
--- a/src/__tests__/cli.internal.test.js
+++ b/src/__tests__/cli.internal.test.js
@@ -112,6 +112,12 @@ describe('formatValue', () => {
     expect(formatValue(5000)).toBe('5.00K');
     expect(formatValue(-1500)).toBe('-1.50K');
   });
+  it('should promote K to M at the rounding boundary', () => {
+    expect(formatValue(999999.994)).toBe('999.99K');
+    expect(formatValue(999999.995)).toBe('1.00M');
+    expect(formatValue(999999.999)).toBe('1.00M');
+    expect(formatValue(-999999.995)).toBe('-1.00M');
+  });
 
   it('should format integers without decimals', () => {
     expect(formatValue(42)).toBe('42');
--- a/src/__tests__/cli.internal.test.js
+++ b/src/__tests__/cli.internal.test.js
@@ -113,7 +113,7 @@ describe('formatValue', () => {
     expect(formatValue(-1500)).toBe('-1.50K');
   });
   it('should promote K to M at the rounding boundary', () => {
-    expect(formatValue(999999.994)).toBe('999.99K');
+    expect(formatValue(999990)).toBe('999.99K');
     expect(formatValue(999999.995)).toBe('1.00M');
     expect(formatValue(999999.999)).toBe('1.00M');
     expect(formatValue(-999999.995)).toBe('-1.00M');
--- a/src/cli.js
+++ b/src/cli.js
@@ -231,7 +231,11 @@ export function formatValue(val) {
   if (val === null || val === undefined) return '';
   if (typeof val === 'number') {
     if (Math.abs(val) >= 1000000) return (val / 1000000).toFixed(2) + 'M';
-    if (Math.abs(val) >= 1000) return (val / 1000).toFixed(2) + 'K';
+    if (Math.abs(val) >= 1000) {
+      const formatted = (val / 1000).toFixed(2);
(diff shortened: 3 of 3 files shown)


kome12:
@dolmaciabdullah-byte Can you address the feedback raised by the pr-reviewer?

dolmaciabdullah-byte:
@kome12 Thanks for the review!

Addressing the pr-reviewer findings:

1. **Unit test coverage (Medium)**: The boundary behavior can be verified with the existing `formatValue` tests in `cli.internal.test.js`. Happy to add explicit test cases for `999999.995`, `999999.999`, and `-999999.995` if you'd like let me know.

2. **Same pattern could apply to other tiers (Low)**: Currently K and M are the only two tiers, so the boundary issue only exists at K→M. If additional tiers are added in the future (e.g. B for billions), the same promotion pattern should be applied. For now, the fix is scoped to the existing code.

Let me know if you'd like me to add the test cases!

kome12:
> @kome12 Thanks for the review!
> 
> Addressing the pr-reviewer findings:
> 
> 1. **Unit test coverage (Medium)**: The boundary behavior can be verified with the existing `formatValue` tests in `cli.internal.test.js`. Happy to add explicit test cases for `999999.995`, `999999.999`, and `-999999.995` if you'd like let me know.
> 2. **Same pattern could apply to other tiers (Low)**: Currently K and M are the only two tiers, so the boundary issue only exists at K→M. If additional tiers are added in the future (e.g. B for billions), the same promotion pattern should be applied. For now, the fix is scoped to the existing code.
> 
> Let me know if you'd like me to add the test cases!

@dolmaciabdullah-byte Thanks! The fix itself looks correct. Two things before we merge:
1. Please add the regression tests: If you can add the three boundary cases from the PR description (999999.995, 999999.999, -999999.995) to the existing `formatValue` block in `cli.internal.test.js`. Worth also pinning 999999.994 → 999.99K so the lower boundary is locked in too. 
2. Changeset: this changes user-visible output (1000.00K → 1.00M), so per `CLAUDE.md` it needs a `.changeset/*.md` with a patch bump. 

Once the tests + changeset are in, this is good to go.

dolmaciabdullah-byte:
Addressed both points in a7ab05c:

1. Changeset added .changeset/fix-formatvalue-km-boundary.md, patch bump.
2. Four boundary tests added: 999999.994 → 999.99K, 999999.995 → 1.00M, 999999.999 → 1.00M, -999999.995 → -1.00M.

Thanks @kome12.

dolmaciabdullah-byte:
Fixed the failing test in 219e8f8 999999.994 / 1000 = 999.999994, which .toFixed(2) rounds to 1000.00, so the fix correctly promotes it to 1.00M. Replaced with 999990 which genuinely stays at 999.99K.

kome12 (approved):
@dolmaciabdullah-byte Thanks for the fix!

dolmaciabdullah-byte:
Thank you @kome12

PR #527, opened 2026-08-27
Bound source-chain outflow on cross-chain bridge swaps

kome12:
## Summary

Cross-chain (bridge) swaps previously skipped swap-outcome verification entirely — the only checks left were metadata-level (`assertQuoteMatchesRequest`) and static instruction inspection, neither of which bounds the actual amount that leaves the wallet. A compromised or buggy bridge quote could move far more than the declared input with nothing to catch it.

- `assertSwapOutcome` (EVM) and `assertSolanaSwapOutcome` (Solana) now derive `isBridge` from the persisted request intent (`request.toChain !== request.chain`) and skip only the output-arrival *delta comparison* for a bridge — the output settles on the destination chain and can never appear in a source-chain simulation.
- The input-outflow cap and no-sibling-drain assertions (plus EVM's NFT/approval checks) now run unconditionally, including for bridges.
- Bridges additionally require a **positive input outflow** — a no-op transaction (empty deltas) no longer verifies as a successful bridge — and still run the quote's **output-amount integrity checks** (present / integer / positive); only the source-chain delta comparison is skipped.
- `verifySwapOutcome`/`verifySolanaSwapOutcome` no longer early-return for bridges, so the simulation actually runs; they log an info line when output verification was skipped (the native-SOL input line no longer contradicts it).

## Scope of the guarantee

This bounds the **source-chain outflow** — how much leaves the wallet on the origin chain — and forbids any non-input token or NFT leaving. It does **not** prove the destination recipient or the eventual destination-chain output; a source-chain simulation structurally cannot. The recipient is AML-screened separately immediately before signing, but its correctness is not re-derived from the simulated deltas. Follow-up ticket to track destination-side attestation.

Went with the strict design (enforce the sibling-drain check on bridges too, not just the outflow cap).

## Test plan

- [x] `npm test` (2258 tests) and `npm run lint` green
- [x] Unit tests for both chains: bridge passes with outflow+sibling checks enforced and output-delta skipped; still blocks an outflow-cap breach; still blocks a sibling drain; **rejects a no-op (zero-outflow) bridge**; **rejects a missing/zero output-amount quote**; native-SOL bridge log line asserted non-contradictory; same-chain behavior unchanged
- [x] Real-money e2e on mainnet (`bridge-test` wallet), both directions, native + token input — all four routes passed with no false-blocks:
  - Solana SOL (native) → Base USDC
  - Solana USDC (token) → Base USDC
  - Base ETH (native) → Solana SOL
  - Base USDC (token, incl. an ERC-20 approval) → Solana USDC

### Not yet exercised: swap-then-bridge through a wallet-owned intermediate

The strict design is most likely to false-block a route that source-swaps through an intermediate token held by the wallet (the sibling-drain check would fire on that token leaving). No such route was exercised end-to-end: the attempt failed to compile for an unrelated, pre-existing reason (no address-lookup-table support yet in the Solana instruction compiler) before ever reaching this verifier. For mainstream bridge aggregators the source swap happens inside the router contract, so intermediate tokens never leave the user's wallet and the check should not fire — but this is unproven, so treat the four passing bridge e2es as evidence for the simple routes only. Note the failure mode here is **fail-closed** (reject → fall through to the next quote), not a drain. Filing the ALT-support gap separately as the gating follow-up.


kome12:
@Codier This is for the cross chain bridge swap that was raised in the #513 PR. Might be best if you review as you raised the issue.

src/trade-validation.js
```diff
@@ -1005,9 +1025,50 @@ export function assertSwapOutcome(request, quote, sim, { slippage, expectedSpend
...
+  // A bridge drops assertion 2 (its output settles on the destination chain),
+  // and for a normal swap that positive-output check is what implicitly proves
+  // the input was actually consumed. Restore that with an intent-relative LOWER
+  // bound on the source-chain outflow: a bare `outflow > 0` is too weak — a
+  // fee-only or 1-unit no-op would still verify a bridge that never funded its
+  // input. For exactIn the outflow must be ~the requested input (assertion 1
+  // already caps it above); for exactOut the input is variable up to the cap, so
+  // only a positive-outflow floor is meaningful. EVM native input delta is the
+  // transferred value with gas excluded, so the outflow is exact — no fee slack.
+  if (isBridge) {
```
Codier, line 1047: **An EVM bridge that pays a native fee via `msg.value` will now be rejected, and bridges never hit this check before.**

Assertion 3 below iterates every non-input delta with `dust = 0n`:

```js
const dust = siblingDustThreshold > 0n ? siblingDustThreshold : 0n;
for (const [token, delta] of Object.entries(deltas)) {
  if (token === inputToken) continue;
  if (delta < 0n && -delta > dust) throw fail(...);
}
```

and the EVM sim does track native movement — `EVM_NATIVE_SENTINEL`, populated from `traceTransfers` synthetic logs or summed `callTracer` frame `value` (`swap-simulation.js`). `verifySwapOutcome` passes no `siblingDustThreshold`, so it's a strict zero.

So on a bridge with a **token** input that also sends native ETH as a protocol/relayer fee, native shows up as a sibling outflow and the quote fails closed. That's not a drain, but it's an availability regression: before this PR bridges returned early and never reached assertion 3 at all.

Worth noting the asymmetry — `assertSolanaSwapOutcome` gives a native sibling `NATIVE_FEE_RENT_SLACK_LAMPORTS` (0.013 SOL) of room for exactly this reason, and EVM gives it nothing.

I don't think the four e2e routes could have caught this: the two native-input routes are exempt via assertion 1, and the Base USDC route evidently doesn't charge a native fee. My question is just whether any route in the aggregator set does. If none do, say so and I'm happy. If some do, a bounded native-fee tolerance for bridges (or threading the quote's declared native fee through as `siblingDustThreshold`) seems like the smaller fix.

This is the same shape as the swap-then-bridge-through-an-intermediate case you flagged — assertion 3 firing on a legitimate bridge — so I'd track them together rather than as two follow-ups.

src/trade-validation.js
```diff
@@ -1252,11 +1323,55 @@ export function assertSolanaSwapOutcome(request, quote, sim, { slippage, sibling
...
+        requested = BigInt(request.amount);
+      } catch {
+        throw fail(`requested input amount (${request.amount}) is not an integer.`);
+      }
+      if (requested <= 0n) throw fail(`bridge exactIn request has a non-positive input amount (${requested}).`);
+      const floorSlack = inputIsNative ? NATIVE_FEE_RENT_SLACK_LAMPORTS : 0n;
+      // Clamp the floor to a positive minimum: for a native bridge smaller than
+      // the fee/rent slack a real leg is indistinguishable from a fee-only no-op,
+      // so the tightest we can still require is a non-zero outflow.
+      const minOutflow = requested > floorSlack ? requested - floorSlack : 1n;
```
Codier, line 1359: **The native-SOL floor is a lot weaker than the changeset suggests, and I think the changeset should say so.**

`NATIVE_FEE_RENT_SLACK_LAMPORTS` is 13,000,000 (0.013 SOL), so:

- requested <= 0.013 SOL → `minOutflow` clamps to `1n`, i.e. effectively no lower bound at all
- requested = 0.02 SOL → floor is 0.007 SOL, so a leg that moves ~35% of the intent still verifies

The comment right above is honest about why, and I agree there's no tighter bound available for a native leg. But the changeset says "a fee-only or partial no-op no longer verifies", which is true for large legs and not true for small ones. I'd just make the changeset match the code.

Codier (commented):
Read the whole thing — both verifier paths, the sim layer, and the tests. This is a real gap being closed and the design is right, so nothing here is a blocker in the fund-safety sense. Three comments, two of them inline.

The call I want to call out as correct: `isBridgeRequest` derives bridge-ness from `request.toChain` on the **persisted intent**, not from the API's quote response. That means a malicious quote can't declare itself a bridge to shed the output-arrival check. That's the crux of this change and it's the right place to read it from.

Restoring the input floor is the other good instinct. Assertion 2 was quietly doing double duty — proving the output arrived *and* implicitly proving the input got consumed — so dropping it for bridges would have let a no-op verify. The intent-relative lower bound is the right substitute.

Two inline comments:

1. An EVM bridge that pays a native fee via `msg.value` alongside a **token** input now trips assertion 3 (strict-zero sibling tolerance), and bridges never reached that check before this PR. Fail-closed, so not a drain, but it can break routes that work today. One question for you inline.
2. The native-SOL floor collapses to 1 lamport below 0.013 SOL — worth reflecting in the changeset.

And a third, changeset-level: bridges now go through the simulation, so a sim **revert** returns `proceed: false` and the quote gets dropped (only `NO_SIM_RPC` / `SIM_RPC_ERROR` / `NOT_SIM_CAPABLE` degrade to proceed). That's the correct behaviour, but it's a new failure mode for bridges specifically and I'd name it in the changeset next to the rest.

Tests are genuinely good — 25 new cases, symmetric across both chains, covering no-op rejection, partial outflow, both exactOut floors, output-integrity-still-enforced, unrecognized swap mode failing closed, and same-chain behaviour unchanged. Validating `swapMode` in the pre-signing intent checks is the right belt-and-braces since it runs even under `--no-verify-outcome`.

Holding my approval only on comment 1 — one answer about whether any route charges a native bridge fee and I'll approve. Not requesting changes; I don't want to sit on a PR that closes a real hole over a question.

Codier (approved):
Hey kome12, I've checked the latest commit and the e2e results in the PR body.

The Base USDC -> Solana USDC pass shows that the native-fee concern isn't a problem for the current routes, which settles my main question.

The updated changeset correctly covers the native-SOL floor and the fail-closed bridge revert mode. This closes a major gap by binding bridge outcomes to the persisted intent. Tests look solid.

Approving.

src/trade-validation.js
kome12: Good catch and confirmed. It's fail-closed (drop the quote, no drain), and the tested Base USDC route doesn't charge a native fee, but I can't guarantee no route in the aggregator set does. Rather than rush a bound here (allowing up to tx.value would just reopen the native-ETH drain this check closes), I'll take it as a separate follow-up along with the swap-then-bridge-through-an-intermediate case. Merging this as the strict improvement over the previous skip-entirely behavior.

PR #506, opened 2026-08-19
API-319: Reconcile MCP setup docs — canonical endpoint, API-key URL, and per-client paths

gulshngill:
## Why

MCP setup guidance is inconsistent across surfaces: the README had no MCP section at all, and the public MCP docs use a generic-client snippet that is awkward against the hosted endpoint (`--allow-http` on an HTTPS URL, and a `--header "NANSEN-API-KEY: "` form that mcp-remote mis-splits on the space).

This PR establishes the canonical values in the repo and gives every client one unambiguous path.

## Canonical values

- **MCP endpoint:** `https://mcp.nansen.ai/ra/mcp` (streamable HTTP, `NANSEN-API-KEY` header)
- **API-key URL (auth / MCP setup only):** `https://app.nansen.ai/auth/agent-setup`
- **Credit top-up URL (credit warnings only):** `https://app.nansen.ai/api?tab=api` — the credits tab of the billing page

## What changed

- **README — new `## MCP` section** with the canonical endpoint/auth/key URL and per-client paths:
  - *Claude Desktop and Cursor:* pointed at [docs.nansen.ai/mcp/connecting](https://docs.nansen.ai/mcp/connecting), where the `.dxt` bundle and the Cursor install deep link are maintained. The README does not duplicate those install artifacts and does not claim either is one-click from here.
  - *One-command:* Claude Code via `claude mcp add --transport http nansen … --header "NANSEN-API-KEY: <your-key>"`
  - *Manual:* native `url` + `headers` JSON for any streamable-HTTP client, and a corrected `mcp-remote` bridge for stdio-only clients (no `--allow-http`, header passed as a single `NANSEN-API-KEY:${NANSEN_API_KEY}` arg with the key in `env`)
  - *Claude Tag (Claude in Slack):* admin-configured path (plugin `.mcp.json` + host credential for `mcp.nansen.ai`), with an explicit per-user fallback to Claude Code / Claude Desktop
- **Credit warnings — README error table + `src/response-meta.js`:** the out-of-credits and low-credit warnings move from the bare `app.nansen.ai/api` to the credits tab of that same billing page, `app.nansen.ai/api?tab=api`. `auth/agent-setup` is not used here — it stays confined to auth and MCP setup.
- **Tests:** `creditWarning` assertions pin the top-up URL and assert `auth/agent-setup` does not leak into either warning
- **Changeset:** patch

## Follow-up for the docs site (not in this repo)

[...]

PR #529, opened 2026-08-27
API-317: follow up on the retired Cursor deep link in the MCP README

gulshngill:
## Why

nansen-api#1803 removes the Cursor install deep link from [docs.nansen.ai/mcp/connecting](https://docs.nansen.ai/mcp/connecting). This README — already published on npm (1.41.1) — tells readers that link is *maintained* on that page:

> **Claude Desktop and Cursor:** setup instructions for both — the Claude Desktop `.dxt` bundle and **the Cursor install deep link** — are in the connection docs: docs.nansen.ai/mcp/connecting.

The moment #1803 merges, that sentence points at an artifact that no longer exists. **This PR should merge first**, so the window never opens.

Note this reverses the follow-up #506 filed for itself ("worth dropping `--allow-http` so the deep link matches"). API-319 wanted the deep link fixed; API-317 concluded it can't be — Cursor persists a deep link's decoded config as-is, so a static link can only ever ship a placeholder key, and a personalized one would put a live key in a URL.

## What changed

- **Drop the artifact enumeration** from the Claude Desktop / Cursor pointer, so the README stays correct as the docs page evolves rather than naming specific install artifacts it doesn't own.
- **Pin `mcp-remote@0.2.1`** (the current release) in the stdio bridge example, replacing `@latest`. The bridge handles the API key on every request, so the version is fixed rather than letting `npx` pull a new release automatically. Bumping is invited — review the release, update the pin.
  - **Correction from review:** an earlier revision of this PR pinned `0.1.38` and justified it as matching what `nansen mcp install` writes and `nansen mcp verify` requires. No such command exists — not on `main` (`schema.json` has no `mcp` key, `mcp-remote` appears only in `README.md`) and not in published 1.40.1. `MCP_REMOTE_PIN = 'mcp-remote@0.1.38'` lives only in #487, which is unmerged and `CONFLICTING`. The `.dxt` is not the source either: it vendors `mcp-remote ^0.1.18` and never invokes `npx`. `0.1.38` is ~7 months and ~13 releases behind current.
  - `0.2.1` verified as a drop-in: identical header parser, identical `${VAR}` substitution, identical `allowHttp` guard, connects to prod with the key substituted from the environment.
- **Correct the header-formatting note.** It said "keep the header as one argument with no space after the colon". The no-space part is wrong: `mcp-remote` parses headers with `/^([A-Za-z0-9_-]+):\s*(.*)$/`, so whitespace after the colon is trimmed. The note now says what actually matters — name and value in a single `args` entry, and the key in `env` rather than the argument list.

## Verification

- `mcp-remote@0.1.38` header parser read from source (`package/dist/chunk-65X3S4HB.js:20713`) — confirms the `\s*` trim.
- Env substitution confirmed live at `chunk-65X3S4HB.js:20850-20861` and by running the documented command against prod: `Replacing ${NANSEN_API_KEY} with environment value in header 'NANSEN-API-KEY'` → `Proxy established successfully`. So the `env` form this README already recommended does work — only the rationale sentence was wrong.
- `npm run lint` clean; `npx vitest run src/__tests__/response-meta.test.js` → 44 passed. No source files touched.

## Checklist

- [x] Tests pass
- [x] `src/schema.json` updated if new commands or flags were added — n/a, docs-only
- [x] `README.md` updated
- [x] Changeset added (patch — the README ships to npm)

### DIFF
--- a/.changeset/mcp-readme-cursor-followup.md
+++ b/.changeset/mcp-readme-cursor-followup.md
@@ -1,0 +1,8 @@
+---
+"nansen-cli": patch
+---
+
+Docs: stop pointing at the retired Cursor install deep link, pin the
+`mcp-remote` bridge to the version `mcp install` writes, and correct the
+header-formatting note (whitespace after the colon is trimmed; the key belongs
+in `env`, not in the argument list).
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ claude mcp add --transport http nansen https://mcp.nansen.ai/ra/mcp --header "NA
       "command": "npx",
       "args": [
         "-y",
-        "mcp-remote@0.1.38",
+        "mcp-remote@0.2.1",
         "https://mcp.nansen.ai/ra/mcp",
         "--header",
         "NANSEN-API-KEY:${NANSEN_API_KEY}"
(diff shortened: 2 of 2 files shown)


Codier (dismissed):
Approving.

The question I left on nansen-api#1803 applies to the `mcp-remote@0.1.38` pin here too — I can't find the `nansen mcp install` / `nansen mcp verify` that's meant to set it, and this repo is where it would live. Answer it over there and this is fine as is.

PR #530, opened 2026-08-27
Tolerate a bounded native-token fee on bridge swap-outcome checks

kome12:
## What

Before signing a swap, `nansen trade execute` runs a balance-delta simulation and checks the result against the persisted request intent (defense-in-depth against a bad quote). One of those checks — "no token other than the input leaves the wallet" (assertion 3) — used a strict-zero tolerance for every non-input token, cross-chain bridges included.

Some bridges pay their network fee in the chain's native token (via `msg.value`) on a **token-input** route. That native value leaves the wallet as a *sibling* of the input token, so the strict-zero check rejected the quote and fell through to the next one — potentially failing an otherwise-valid bridge.

## Change

Relax assertion 3 to tolerate a **native-token** sibling outflow, but:
- **only for bridges** — same-chain swaps keep strict-zero for every token;
- **only for the native token** — every ERC-20 sibling stays strict-zero;
- **bounded** by `min(tx.value, EVM_BRIDGE_NATIVE_FEE_SLACK)`, where the cap is 0.002 ETH.

Capping at the fixed ceiling — rather than trusting the quote's declared `tx.value` outright — is what keeps the check from being turned into a native-token drain: a hostile quote can't set `tx.value` to the whole balance and have it waved through (the same `tx.value` is simulated, so a large declared value produces a large simulated outflow that trips the cap). This mirrors the native-only carve-out the Solana path already applies.

Behavior is unchanged for same-chain swaps and for any ERC-20 movement.

## Testing

- Unit tests for both `assertSwapOutcome` and the `verifySwapOutcome` wiring: a native fee within the bound verifies; above the bound blocks; an inflated `tx.value` drain still blocks at the cap; ERC-20 siblings and same-chain swaps stay strict.
- Full suite green, lint clean.
- No live bridge e2e: the native-fee slack is defense-in-depth for a route shape that is not currently reachable through the shipping quote path, so there is no route to exercise it end-to-end. The unit tests carry the proof.

### DIFF
--- a/.changeset/bridge-native-fee-sibling-tolerance.md
+++ b/.changeset/bridge-native-fee-sibling-tolerance.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
+Trade safety: tolerate a bounded native-token fee on cross-chain bridge swaps. The pre-signing swap-outcome check rejected any non-input token leaving the wallet, which could reject a legitimate bridge that pays its network fee in the native token on a token-input route. The tolerance is capped and applies only to the native token on bridges; every other token, and all same-chain swaps, stay strict.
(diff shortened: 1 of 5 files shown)


Codier (approved):
Approved — no changes.

PR #533, opened 2026-08-28
Verify Hyperliquid bridge action against request intent before signing

kome12:
## Summary

The Hyperliquid bridge withdrawal leg (`nansen bridge execute`, HL → base/ethereum/arbitrum) signed a server-supplied EIP-712 action almost verbatim — only the deposit action's EIP-712 domain was pinned client-side. Neither the deposit action's contents nor the relayer's authorize-step typed-data payload were checked against what the user actually requested before signing.

This adds static, pre-signing checks on both legs:

**Deposit (`sendAsset`) action:**
- Action type allowlisted to `sendAsset`
- Amount capped at what the user actually requested (anchored to the amount persisted client-side at quote time, not a server-displayed field), one-sided so it can never false-block a smaller amount
- Target network (`hyperliquidChain`) pinned to `Mainnet`
- Source token pinned to Hyperliquid's canonical USDC identifier; `sourceDex`/`destinationDex`/`fromSubAccount` required empty — the only shape this bridge path is designed to produce

**Relayer authorize (`NonceMapping`) step:**
- EIP-712 domain (name, version, chainId, verifyingContract), primary type, and the exact ordered `{name, type}` field list are all pinned to the real captured shape
- `wallet`/`depositor` required to equal the signing wallet
- The signature's POST target is built entirely from a hardcoded constant and only compared against the server-supplied endpoint — never constructed from it — so there's nothing left to redirect

**Deliberately not bound:** the deposit action's `destination` field is the *relayer's* deposit address, not the user's — binding it to the recipient would reject every legitimate withdrawal. The final payout recipient is held off-chain by the relayer and isn't in anything we sign, so a fully malicious server could still misdirect it; loss is bounded to the amount cap, not eliminated. This residual is documented in code.

## Why three commits

This went through two rounds of independent review after the first commit:
- Round 2 found the authorize step's checks were far too permissive (only checked optional `wallet`/`depositor` value fields, nothing about domain/type shape), and that the deposit action's amount/token/routing fields had similar gaps.
- Round 3 found the round-2 authorize pinning was still partial (domain name only, field names only, not full domain/order/types) — and fixing that properly required getting the *real* EIP-712 shape via a fresh read-only quote capture, which turned up that round 2's guessed domain name (`"Relay"`) didn't match production (real name is `"RelayNonceMapping"`, with different field types/order than guessed). Round 2's code, if shipped, would have refused every legitimate Hyperliquid withdrawal.

## Test plan

- [x] `npm test` — 2328 passed, 2 skipped
- [x] `npm run lint` — clean
- [x] New test file `bridge-hl-action-intent.test.js` (41 tests) covering both the direct assertion functions and end-to-end `execute()` paths (local + Privy signers), including regression tests for every gap found across all three commits
- [x] 4 existing bridge test files updated to the corrected fixture shapes
- [x] Manually verified against the real CLI binary with crafted malicious fixtures (all correctly refused before any signing/broadcast — confirmed no network calls or signatures were produced)
- [x] Verified against a fresh, real, read-only `bridge quote` capture that legitimate quotes still pass all checks (no false-block)
- [ ] No live e2e run required — checks are fully static and unit-covered; a read-only quote capture confirmed the wire shape

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


gulshngill:
Checked these out locally against `fix/bridge-hl-action-intent-binding` — built fixtures for each and ran them through `execute()` on both the local and Privy signer paths. The PR's own suite is green (2345 passed, 2 skipped); these are gaps the new checks don't cover rather than regressions in them.

Two I'd hold the merge on.

---

### 1. The action leg's EIP-712 `primaryType` and `types` are still server-controlled

`src/bridge.js:806-807` (local) and `867-868` (Privy) take both off the wire:

```js
const types = signData.eip712Types || {};
const primaryType = signData.eip712PrimaryType || 'HyperliquidTransaction';
```

The pinned domain doesn't constrain this — `HyperliquidSignTransaction` is the domain for *every* HL user-signed action, which the code comment at `:398-400` already acknowledges. And `hashStruct` only hashes fields named in `types[primaryType]`, so the signed digest covers exactly the fields the server chooses to list.

The consequence is that the new checks and the signature don't overlap. `assertHlBridgeActionIntent` validates `amount`, `token`, `hyperliquidChain` and the dex/sub-account fields on the *action object*, but a response supplying a different valid type list produces a digest that doesn't include any of them. I confirmed this end-to-end: a quote whose action passes **every** check the PR adds — `sendAsset`, `Mainnet`, `amount` under the cap, canonical USDC token, empty `sourceDex`/`destinationDex`/`fromSubAccount` — still yields a signature over a completely different Hyperliquid action, byte-identical to one produced by signing that action's digest directly. An agent-approval-style action is the worst case, since it isn't amount-bounded at all, so the cap this PR adds provides no protection against it.

Worth saying plainly: this is pre-existing behaviour, not something this PR introduces. But it's the same threat model the PR sets out to close, and it undercuts the amount cap that's the headline control here, so it seems in scope.

**Fix:** pin the action leg the way the authorize leg already is — `primaryType` to `HyperliquidTransaction:SendAsset` and an exact, ordered 8-field list. `assertHlBridgeAuthorizeIntent` at `:509-537` is the right model to copy. The same pin is needed at the preflight (`:603-604`), which has the identical `|| 'HyperliquidTransaction'` fallback.

I have a runnable PoC for this — sending it to @kome12 directly rather than attaching it here.

---

### 2. The amount cap rejects a large share of valid default-path withdrawals

`floorHyperliquidUsdcBridgeAmount` is only called inside the `--amount-unit` branch (`src/bridge.js:1104`), so the default base-units path persists the amount unfloored. Relay then formats to USDC's 6dp with round-half-up — which this file documents itself at `:117-120` — and the cap at `:471` compares that against the unfloored 8dp value.

Repro: `nansen bridge quote --from hyperliquid --to base --token USDC --amount 200000050` persists `requestedAmountBaseUnits: "200000050"`. Relay formats `2.000001`. `decimalToScaled` gives `200000100`, which exceeds `200000050 + 1`, so `execute` fails with `AMOUNT_MISMATCH` on a withdrawal the user requested legitimately.

It isn't an edge case. Enumerating all 100 possible values of the two dropped digits: residues **50 through 98 fail — 49 of 100**. Residue 99 survives only on the `+1n` slack, and 0–49 round down. So roughly half of all base-unit amounts a user could pass are unexecutable.

Also flagging that the PR description's "one-sided so it can never false-block a smaller amount" is true for smaller amounts but doesn't hold as a safety argument — the server rounding *up* is exactly the legitimate case that gets blocked.

**Fix:** either apply the flooring on the default path too, or normalise both sides to 6dp before comparing. Worth a regression test at residue 50 specifically, since that's the boundary.

---

### Smaller things

**3. Preflight validates items already marked `complete`.** `preflightHlBridgeSteps` (`:593-608`) has no `status === 'complete'` guard, but both step processors do (`:769`, `:838`). A partially-executed bridge can therefore be blocked from resuming by a check against a step that's already been signed and submitted.

**4. A step item with neither `sign` nor `action` is silently skipped, inconsistently.** Preflight skips it, and `processSignatureStepLocal` (`:799`) has no `else` — so `execute()` resolves successfully having signed and submitted nothing. `processSignatureStepPrivy` throws `Unexpected signature step format` (`:871-873`) for the same input, so the two signers disagree on a malformed quote. Separately, an item with no `data` key at all throws a bare `TypeError` from the destructure rather than a `CommandError`.

**5. The `/authorize` pin is exact-match.** `resolveRelayTargetUrl` (`:562-570`) requires `endpoint === '/authorize'`, so `/authorize?requestId=0x…` or a fully-qualified `https://api.relay.link/authorize` are both refused. I couldn't find evidence Relay ever sends either form here, so this may well be fine as-is — noting it as a fragility if Relay ever adds a query param, not as a live break.

**6. Several wire-shape errors suggest the wrong remedy.** The domain, field-set and endpoint mismatches all end in "Request a new quote", but these compare against hardcoded constants — if Relay bumps `version` to `'3'`, every new quote fails identically and the actual fix is a CLI upgrade. Suggest wording that points at updating instead.

---

### Same class on the other leg — fine as a separate PR

Flagging the connection more than the finding. `processEvmStep` (`:710-728`) validates only `from`, then signs the server-supplied `to`, `data` and `value` as given. That's the same shape as what this PR fixes — a server-supplied payload signed without being bound to what the user asked for — just on the deposit leg (`execution_type: 'evm_transaction'`, base → hyperliquid) rather than the withdrawal leg.

The PR is honestly scoped and I'm not calling this a gap in the change: the description says withdrawal leg and that's exactly what it delivers. But the two legs are branches of the same `bridge execute`, so the net effect is that one of the four supported routes still signs unbound once this lands.

Your call whether that's worth folding in here or tracking separately — I'd lean separate PR, given this one has already been through three review rounds and the deposit leg needs a different binding (calldata decode against the quoted deposit, not an EIP-712 field pin). Mainly want it tracked somewhere other than a comment that scrolls away.

kome12:
> Checked these out locally against `fix/bridge-hl-action-intent-binding` — built fixtures for each and ran them through `execute()` on both the local and Privy signer paths. The PR's own suite is green (2345 passed, 2 skipped); these are gaps the new checks don't cover rather than regressions in them.
> 
> Two I'd hold the merge on.
> 
> ### 1. The action leg's EIP-712 `primaryType` and `types` are still server-controlled
> `src/bridge.js:806-807` (local) and `867-868` (Privy) take both off the wire:
> 
> ```js
> const types = signData.eip712Types || {};
> const primaryType = signData.eip712PrimaryType || 'HyperliquidTransaction';
> ```
> 
> The pinned domain doesn't constrain this — `HyperliquidSignTransaction` is the domain for _every_ HL user-signed action, which the code comment at `:398-400` already acknowledges. And `hashStruct` only hashes fields named in `types[primaryType]`, so the signed digest covers exactly the fields the server chooses to list.
> 
> The consequence is that the new checks and the signature don't overlap. `assertHlBridgeActionIntent` validates `amount`, `token`, `hyperliquidChain` and the dex/sub-account fields on the _action object_, but a response supplying a different valid type list produces a digest that doesn't include any of them. I confirmed this end-to-end: a quote whose action passes **every** check the PR adds — `sendAsset`, `Mainnet`, `amount` under the cap, canonical USDC token, empty `sourceDex`/`destinationDex`/`fromSubAccount` — still yields a signature over a completely different Hyperliquid action, byte-identical to one produced by signing that action's digest directly. An agent-approval-style action is the worst case, since it isn't amount-bounded at all, so the cap this PR adds provides no protection against it.
> 
> Worth saying plainly: this is pre-existing behaviour, not something this PR introduces. But it's the same threat model the PR sets out to close, and it undercuts the amount cap that's the headline control here, so it seems in scope.
> 
> **Fix:** pin the action leg the way the authorize leg already is — `primaryType` to `HyperliquidTransaction:SendAsset` and an exact, ordered 8-field list. `assertHlBridgeAuthorizeIntent` at `:509-537` is the right model to copy. The same pin is needed at the preflight (`:603-604`), which has the identical `|| 'HyperliquidTransaction'` fallback.
> 
> I have a runnable PoC for this — sending it to @kome12 directly rather than attaching it here.
> 
> ### 2. The amount cap rejects a large share of valid default-path withdrawals
> `floorHyperliquidUsdcBridgeAmount` is only called inside the `--amount-unit` branch (`src/bridge.js:1104`), so the default base-units path persists the amount unfloored. Relay then formats to USDC's 6dp with round-half-up — which this file documents itself at `:117-120` — and the cap at `:471` compares that against the unfloored 8dp value.
> 
> Repro: `nansen bridge quote --from hyperliquid --to base --token USDC --amount 200000050` persists `requestedAmountBaseUnits: "200000050"`. Relay formats `2.000001`. `decimalToScaled` gives `200000100`, which exceeds `200000050 + 1`, so `execute` fails with `AMOUNT_MISMATCH` on a withdrawal the user requested legitimately.
> 
> It isn't an edge case. Enumerating all 100 possible values of the two dropped digits: residues **50 through 98 fail — 49 of 100**. Residue 99 survives only on the `+1n` slack, and 0–49 round down. So roughly half of all base-unit amounts a user could pass are unexecutable.
> 
> Also flagging that the PR description's "one-sided so it can never false-block a smaller amount" is true for smaller amounts but doesn't hold as a safety argument — the server rounding _up_ is exactly the legitimate case that gets blocked.
> 
> **Fix:** either apply the flooring on the default path too, or normalise both sides to 6dp before comparing. Worth a regression test at residue 50 specifically, since that's the boundary.
> 
> ### Smaller things
> **3. Preflight validates items already marked `complete`.** `preflightHlBridgeSteps` (`:593-608`) has no `status === 'complete'` guard, but both step processors do (`:769`, `:838`). A partially-executed bridge can therefore be blocked from resuming by a check against a step that's already been signed and submitted.
> 
> **4. A step item with neither `sign` nor `action` is silently skipped, inconsistently.** Preflight skips it, and `processSignatureStepLocal` (`:799`) has no `else` — so `execute()` resolves successfully having signed and submitted nothing. `processSignatureStepPrivy` throws `Unexpected signature step format` (`:871-873`) for the same input, so the two signers disagree on a malformed quote. Separately, an item with no `data` key at all throws a bare `TypeError` from the destructure rather than a `CommandError`.
> 
> **5. The `/authorize` pin is exact-match.** `resolveRelayTargetUrl` (`:562-570`) requires `endpoint === '/authorize'`, so `/authorize?requestId=0x…` or a fully-qualified `https://api.relay.link/authorize` are both refused. I couldn't find evidence Relay ever sends either form here, so this may well be fine as-is — noting it as a fragility if Relay ever adds a query param, not as a live break.
> 
> **6. Several wire-shape errors suggest the wrong remedy.** The domain, field-set and endpoint mismatches all end in "Request a new quote", but these compare against hardcoded constants — if Relay bumps `version` to `'3'`, every new quote fails identically and the actual fix is a CLI upgrade. Suggest wording that points at updating instead.
> 
> ### Same class on the other leg — fine as a separate PR
> Flagging the connection more than the finding. `processEvmStep` (`:710-728`) validates only `from`, then signs the server-supplied `to`, `data` and `value` as given. That's the same shape as what this PR fixes — a server-supplied payload signed without being bound to what the user asked for — just on the deposit leg (`execution_type: 'evm_transaction'`, base → hyperliquid) rather than the withdrawal leg.
> 
> The PR is honestly scoped and I'm not calling this a gap in the change: the description says withdrawal leg and that's exactly what it delivers. But the two legs are branches of the same `bridge execute`, so the net effect is that one of the four supported routes still signs unbound once this lands.
> 
> Your call whether that's worth folding in here or tracking separately — I'd lean separate PR, given this one has already been through three review rounds and the deposit leg needs a different binding (calldata decode against the quoted deposit, not an EIP-712 field pin). Mainly want it tracked somewhere other than a comment that scrolls away.

@gulshngill Addressed your feedback. Will handle EVM deposit leg in another PR

gulshngill:
Re-checked against 643b301 — I re-ran the original repros unchanged rather than reading the diff, plus a few cases that weren't in the first pass. Both blockers are closed. Full suite green (2352 passed, 2 skipped), lint clean.

**1 — deposit-leg EIP-712 shape.** Closed. `HL_SENDASSET_PRIMARY_TYPE` and the ordered 8-field list are pinned at the preflight and on both signer paths, with the `|| 'HyperliquidTransaction'` fallbacks gone. The shape-swapped quote is now refused on the local path with nothing reaching `api.hyperliquid.xyz`, and on the Privy path before `ethSignTypedDataV4` is called at all.

Two cases I added beyond the ones in the new tests, both correctly refused:
- a **field reorder** — the same 8 fields with `destination` and `amount` transposed, which changes the struct hash while looking superficially valid
- a **truncated list** with `amount` removed, i.e. the minimal version of the original problem rather than a whole different action shape

And the genuine captured shape still executes end-to-end, so the pin isn't over-tight.

**2 — default-path flooring.** Closed. `--amount 200000050` now persists as `200000000`. I ran residues 0, 49, 50, 51, 98 and 99 end-to-end through `execute()` — all six complete, including 50, which was the boundary that failed before. Also confirmed `base → hyperliquid` is unaffected (`2000050` persists intact), so the floor didn't leak into the EVM direction.

**3, 4, 6.** All closed. Preflight skips `complete` items so a resumed bridge isn't wedged; malformed items raise a `CommandError` up front instead of a bare `TypeError` from the destructure; the local processor throws on the unknown-shape case rather than reporting success having signed nothing; shape-pin errors now point at updating the CLI.

**5** left as-is, which I agree with — that was a fragility note about Relay potentially adding a query param, not a defect.

---

Three minor observations, none blocking:

- The floor is silent — `--amount 200000050` sends `200000000` with no notice. It's 5e-7 USDC and `--amount-unit` already behaved this way, so it's consistent; just noting it's an unannounced adjustment.
- Quotes saved by a pre-fix CLI still carry the unfloored value and will still hit the cap at execute time. The error tells you to request a new quote, which is now exactly the right remedy, so this self-heals.
- The floor is gated on `isBridgeUsdc`, which matches the resolved USDC sentinel address. A raw non-symbol token string on the default path wouldn't floor — but `assertHlBridgeActionIntent`'s token pin rejects non-canonical tokens anyway, so I don't think it's reachable as a false block.

LGTM. The fix mirrors the existing authorize-leg pin rather than inventing a second pattern, which makes the two legs read consistently.

The EVM deposit leg (`processEvmStep`) is still worth its own PR, as discussed above — nothing for this one.

gulshngill (approved):
some non-blocking findings but other than that, it looks good!

PR #534, opened 2026-08-28
Harden login credential handling

gulshngill:
## Summary

- Suppress upstream verification text so raw or transformed credentials cannot reach CLI output.
- Reject redirects for requests carrying API or payment credentials — **now across every credentialed fetch path, not just `api.js`**.
- Keep interactive key **and wallet-password** input masked when stdout is redirected.
- Prefer portable login guidance that avoids putting literal keys in shell history.
- Send invalid-key remediation to the key-management view.

## What changed (follow-up commit `ebb6150`)

The first pass hardened the two `api.js` fetch sites and the login verification message. Review found the same leak class left open elsewhere: undici forwards custom credential headers (`apikey`, `X-API-Key`, `NANSEN-API-KEY`, `privy-app-id`) across a cross-origin redirect — confirmed empirically with a 307 loopback that delivered the `apikey` header to a second host. Only `Authorization` is stripped. The follow-up closes the gaps:

- **`redirect: 'error'` on every remaining credentialed fetch:** `agent` (apikey), `limit-order` (JWT + X-API-Key), `mcp verify` (NANSEN-API-KEY), `privy` (Basic auth), and `swap-simulation` (apikey, Nansen-hosted endpoints only).
- **Wallet password masking:** `wallet.js` `promptPassword` now gates on `stdin.isTTY` instead of `stdout.isTTY`, so a redirected stdout (`wallet export > backup.json`) no longer falls through to readline and echoes the password in cleartext. Prompt/mask characters go to stderr. Exported with injectable streams so the masking is unit-tested.
- **Verification error signal:** `VERIFICATION_FAILED` now branches on the structured `error.code` (rate-limit / server / timeout) instead of collapsing every non-auth failure to a misleading "check your key" — without re-introducing the `error.message` relay.
- **Docs consistency:** `schema.json` `mcp verify` examples lead with the saved-key / `NANSEN_API_KEY` form (`examples[0]` renders verbatim in `--help`); `--api-key` kept only as an explicit override. The `NOT_A_TTY` login hint now leads with `NANSEN_API_KEY` over the history-recording `--api-key`.

## Verification

- `npm test`: 59 files passed; **2,422** tests passed; 2 expected skips.
- `npm run lint`: passed.
- New/updated tests: redirect-refusal assertions on each patched fetch site, wallet-password masking (no cleartext echo when stdout is redirected), and a rate-limit signal test that proves the message is not relayed.
- Live loopback probes: a redirect no longer relays credentials to the redirect target on the `api.js` and `agent` paths; a legitimate quote/verify round-trip is unaffected.

### DIFF
--- a/.changeset/login-key-hygiene.md
+++ b/.changeset/login-key-hygiene.md
@@ -2,6 +2,10 @@
 "nansen-cli": patch
 ---
 
-`nansen login` hygiene: verification failures cannot relay the API key,
-authenticated redirects are rejected, invalid-key remediation points at key
-management, and login guidance leads with paths that avoid shell history.
+Credential hygiene: `nansen login` verification failures cannot relay the API
+key, invalid-key remediation points at key management, and login guidance leads
+with paths that avoid shell history. Every request that carries a credential —
+API key, agent, limit-order JWT/X-API-Key, MCP verify, and Privy auth — now
+refuses to follow HTTP redirects, so a credential can't be relayed to a redirect
+target. Interactive password and API-key prompts stay masked (no cleartext echo)
+even when stdout is redirected.
--- a/.changeset/login-key-hygiene.md
+++ b/.changeset/login-key-hygiene.md
@@ -1,0 +1,7 @@
+---
+"nansen-cli": patch
+---
+
+`nansen login` hygiene: verification failures cannot relay the API key,
+authenticated redirects are rejected, invalid-key remediation points at key
+management, and login guidance leads with paths that avoid shell history.
--- a/README.md
+++ b/README.md
@@ -20,11 +20,11 @@ Three options — pick whichever fits your setup:
 
 1. **API key** (subscription):
    ```bash
-   nansen login --api-key <key>   # save key to ~/.nansen/config.json
-   nansen login --human           # interactive prompt
-   export NANSEN_API_KEY=...      # env var (highest priority)
-   nansen logout                  # remove saved key
+   nansen login --human   # interactive prompt; saves to ~/.nansen/config.json
+   nansen login           # uses NANSEN_API_KEY when already set
+   nansen logout          # remove saved key
    ```
+   For automation, inject `NANSEN_API_KEY` through your environment or secret manager.
    Get your API key at [app.nansen.ai/auth/agent-setup](https://app.nansen.ai/auth/agent-setup).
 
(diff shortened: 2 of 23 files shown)


kome12 (commented):
Some feedback:
1. (P2 - should fix) `schema.json` mcp verify examples still lead with --api-key <key>
src/schema.json:1987-1989. All three examples use the shell-history-unsafe form. `nansen mcp verify --help` renders examples[0] verbatim, so the help output still shows Example: `npx -y nansen-cli mcp verify --api-key <key>`, contradicting this PR's own README change. The docs sweep missed this one file.
→ Reorder so a saved-key / NANSEN_API_KEY example is first; keep --api-key only as an explicit override.

2. (Low - worth fixing for consistency) `swap-simulation.js` credentialed fetch has no redirect: 'error' src/swap-simulation.js:214 (postSim) attaches the apikey header but still follows redirects which is the exact vector this PR closes on `api.js`. Custom apikey headers are not stripped by undici on cross-origin redirects. Gated by `isNansenHostedUrl` so severity is low, but it violates the PR's own threat model.
→ Add redirect: 'error' to that fetch to match api.js.

3. (Low - optional) VERIFICATION_FAILED collapses all non-auth errors to one generic message `src/cli.js `. Dropping `error.message` fixed the leak, but now a rate-limit or 5xx during `getAccount()` shows a misleading "Could not verify API key. Check your internet connection."
→ Branch on error.code (safe, structured) to restore signal without re-introducing the message leak.

gulshngill:
@nansen-pr-reviewer re-review

The single medium finding (unconditional `redirect: 'error'` in `postSim`) is fixed in `00fae55`: the guard is now `redirect: sendApiKey ? 'error' : 'follow'`, so anonymous third-party `NANSEN_BASE_SIM_RPC` calls follow redirects as before, and code/comment now agree. Tests pin both directions; full suite green (2422), lint clean.

gulshngill:
@nansen-pr-reviewer re-review

Follow-up findings addressed in `e7b2392`:
- `cli.js prompt()` masked output now defaults to stderr (matches `wallet.js promptPassword`), so masks never land in a redirected stdout.
- `privy.js _request` and `limit-order.js loFetch` now wrap `fetch` in try/catch, converting a redirect's bare TypeError into an actionable/coded error — same as `api.js`/`agent.js`.

Swept the remaining `redirect: 'error'` sites: `swap-simulation` (catch → SwapSimulationError) and `mcp-verify` (caller → failed check) already degrade cleanly, and no credentialed fetch site is left following redirects. Full suite green (2425), lint clean.

kome12 (approved):

PR #535, opened 2026-08-28
fix(trade): fall back to gas estimate when EVM quotes omit gas limits

crazywriter1:
## Summary

When an EVM swap quote has no gas fields (`quote.gas` and `tx.gas`/`gasLimit` both zero/missing), WalletConnect and local-wallet execute paths could send or sign with `gas: "0"`. Privy already handled this; WC and local did not.

- Add shared `resolveEvmSwapGasLimit()` (estimate ×1.5, then 210000 fallback)
- Use it on Privy, WalletConnect, and local EVM signing paths
- Regression tests for native/ERC-20, bridge, RPC failure, and happy path

## Test plan

- [x] `npm test` — 2300 passed
- [x] `npm run lint`
- [x] Added tests for zero/missing gas quotes on WalletConnect, local wallet, Privy, ERC-20 swaps, LiFi bridge, and RPC estimate failure (210000 fallback)
- [ ] `npm run test:trade` — skipped (no funded wallet)

### DIFF
--- a/.changeset/evm-swap-gas-zero-fallback.md
+++ b/.changeset/evm-swap-gas-zero-fallback.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
+Fix EVM swap execution when quotes omit gas limits: WalletConnect and local wallet paths now fall back to eth_estimateGas (×1.5) and then 210000, matching the Privy path.
--- a/src/__tests__/trading.test.js
+++ b/src/__tests__/trading.test.js
@@ -701,6 +701,13 @@ describe('resolveEvmSwapGasLimit', () => {
     expect(await resolveEvmSwapGasLimit(quote, { chain: 'base', from: '0xabc' })).toBe(250000);
   });
 
+  it('parses hex gas without triggering estimation', async () => {
+    vi.stubGlobal('fetch', vi.fn(() => Promise.reject(new Error('should not estimate'))));
+    const quote = { transaction: { to: LIFI_ROUTER, data: '0x', gas: '0x88530' } };
+    expect(await resolveEvmSwapGasLimit(quote, { chain: 'base', from: '0xabc' })).toBe(558384);
+    vi.unstubAllGlobals();
+  });
+
   it('estimates with a 1.5x buffer when both are zero', async () => {
     vi.stubGlobal('fetch', vi.fn().mockResolvedValue({
       ok: true,
--- a/src/__tests__/trading.test.js
+++ b/src/__tests__/trading.test.js
@@ -1536,12 +1536,14 @@ describe('WalletConnect execute support', () => {
       }),
     });
 
-    const cmds = buildTradingCommands({ log: () => {}, exit: () => {} });
+    const logs = [];
+    const cmds = buildTradingCommands({ log: (msg) => logs.push(msg), exit: () => {} });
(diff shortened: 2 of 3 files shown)


crazywriter1:
@nansen-pr-reviewer re-review

crazywriter1:
@nansen-pr-reviewer re-review

crazywriter1:
> ## pr-reviewer Summary for #eee76dc
> 📝 **1 finding**
> 
> Review completed. Please address the findings below.
> 
> ### Findings by Severity
> Severity	Count
> 🟡 Medium	1
> **Review effort:** 2/5 (Simple)
> 
> ### Summary
> This is a clean, well-targeted bug fix. The new `resolveEvmSwapGasLimit()` helper correctly handles decimal gas strings, hex gas strings, zero/missing gas, and RPC failure, and all three signing paths (Privy, WalletConnect, local) now use the shared implementation. Changeset is present and correctly typed as `patch`. Tests mock all RPC calls and do not hit real networks. One medium-severity issue is worth addressing.
> 
> Findings
> ### `src/trading.js` — `logEvmSwapGasResolution` silently drops logs when `apiGas === txGas > 0` (medium)
> **Severity:** medium — incorrect behavior (silent log suppression), but does not affect gas resolution itself.
> 
> ```js
> // Line 1136
> if (apiGas > 0 && finalGas !== txGas) {
>   log(`  ℹ Using API gas ${finalGas} (tx.gas was ${txGas})`);
> }
> ```
> 
> When the API gas and `tx.gas` agree (e.g. `quote.gas = "210000"`, `tx.gas = "210000"`), `finalGas === txGas` and **nothing is logged**. This is harmless when they match, but it also means the "API gas in use" path is never confirmed in the logs for what is arguably the most common happy-path case. The condition was ported from the old Privy-only implementation, but now it serves all three paths — a mismatch between the intent ("log when API gas is used") and the actual behavior ("log only when API gas overrides a different tx.gas") is more visible in the shared helper.
> 
> **Suggested fix:** Change the condition to log whenever the API value is in use, regardless of whether it equals `tx.gas`:
> 
> ```js
> if (apiGas > 0) {
>   // Only log something noteworthy: when api and tx disagree, or when there is no tx.gas at all
>   if (finalGas !== txGas) {
>     log(`  ℹ Using API gas ${finalGas} (tx.gas was ${txGas})`);
>   }
> } else if (txGas === 0) {
>   log(`  ℹ Using estimated gas ${finalGas} (quote had no gas)`);
> }
> ```
> 
> Or simply drop the `finalGas !== txGas` guard if matching values should also be logged. The current behavior is arguably intentional (avoid noise when values agree), but if so, add a comment explaining the design choice so future reviewers don't need to reconstruct it.
> 
> _Token usage: 173 input, 5,893 output, 607,150 cache read, 45,940 cache write | [Usage Guide](https://github.com/nansen-ai/webhook-botter/blob/main/docs/usage/pr-reviewer.md)_
> 
> _New pushes are reviewed automatically with a 10-minute cooldown between reviews. To request a review at any time, comment `@nansen-pr-reviewer re-review`._

Intentional. Not a bug.

Only unusual cases are logged:
- API gas ≠ tx.gas
- both missing → estimation/fallback

When quote.gas and tx.gas already match (the common happy path), logging every swap would just be noise. Gas resolution itself is unchanged. @kome12

PR #538, opened 2026-08-31
Harden the Hyperliquid bridge EVM deposit leg against tampered calldata

kome12:
## Problem

The `nansen bridge` **EVM deposit leg** (Base → Hyperliquid via Relay) is a separate execution path from `nansen trade execute` and never received the trade path's hardening. `processEvmStep` signed the server-supplied `approve` and `deposit` transactions almost verbatim — the only validation was a `from` check. A compromised API (or the Relay upstream it proxies) could return calldata for `approve(attacker, MAX_UINT256)` or an arbitrary drain call and the CLI would sign it with the user's local key.

## What changed

A new `assertEvmBridgeStepIntent`, run as a **preflight over every step before any is broadcast**, binds each transaction to the user's intent. The pinned constants were captured from real read-only `bridge quote` responses (confirmed stable across three quotes), not guessed.

- **Approvals are re-scoped** through the existing hardened `encodeApproveCalldata`: the approve must target the origin-chain USDC contract, the spender must be the known Relay router, and the amount is capped to the requested input (rejects `MAX_UINT256`/unbounded).
- **The deposit target is pinned**: `to` + selector must match the known Relay router and deposit method; the decoded args are bound (depositor == signer, token == USDC, amount ≤ requested), and native `value` must be zero.
- **The plan is bounded**: at most one approve and one deposit. ERC-20 `approve` overwrites (not increments) the allowance, so without this a tampered response could chain repeated `[approve(requested), deposit(requested)]` pairs — each passing every per-item check — to pull the requested amount N times and drain the whole balance. (Caught in security review.)

The overall guarantee: **at most the requested amount can leave the wallet for a given quote.**

### Known residual
The Relay payout recipient is held off-chain behind the request id and does not appear in the signed bytes (the deposit's `id` arg is opaque), so full recipient binding isn't possible on this leg — the same structural limit as the sibling withdrawal-leg work. Loss is bounded by the checks above.

## Testing

- New `bridge-evm-deposit-intent.test.js` (46 assertions): approve/deposit binding, MAX rejection, over-cap, wrong token, depositor redirect, non-zero value, malformed calldata, and the plan-level amplification bound.
- Existing bridge fixtures updated for the new preflight.
- Full suite green; lint clean.
- **Real-money round-trip e2e passed**: Base → HL deposit of 2 USDC (exercises the new guard) and HL → Base withdrawal back; Base balance reconciled exactly, confirming the guard does not false-block a legitimate deposit.

### DIFF
--- a/.changeset/harden-bridge-evm-deposit.md
+++ b/.changeset/harden-bridge-evm-deposit.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
+Harden the Hyperliquid bridge deposit leg: EVM approvals are now re-scoped to the requested amount (never unlimited) and the deposit target contract/method is pinned, so a tampered quote can't drain the wallet.
--- a/src/__tests__/bridge-evm-deposit-intent.test.js
+++ b/src/__tests__/bridge-evm-deposit-intent.test.js
@@ -171,6 +171,57 @@ describe('assertEvmBridgeStepIntent — cross-cutting', () => {
       expect(e.code).toBe('INVALID_INPUT');
     }
   });
+
+  it('refuses a deposit whose amount word is not valid hex, instead of throwing a raw SyntaxError', () => {
+    const badAmount = '0xe8017952' + word(SIGNER) + word(USDC) + 'zz'.repeat(32) + word('a');
+    const txData = { to: ROUTER, data: badAmount, value: '0' };
+    expect(() => assertEvmBridgeStepIntent(txData, intent)).toThrow(/malformed deposit calldata/);
+    try {
+      assertEvmBridgeStepIntent(txData, intent);
+    } catch (e) {
+      expect(e.code).toBe('INVALID_INPUT');
+    }
+  });
+
+  it('refuses an approve whose amount word is not valid hex, instead of throwing a raw SyntaxError', () => {
+    const txData = { to: USDC, data: '0x095ea7b3' + word(ROUTER) + 'zz'.repeat(32), value: '0' };
+    expect(() => assertEvmBridgeStepIntent(txData, intent)).toThrow(/malformed approve calldata/);
+    try {
+      assertEvmBridgeStepIntent(txData, intent);
+    } catch (e) {
+      expect(e.code).toBe('INVALID_INPUT');
+    }
+  });
+
+  it('refuses an unparseable native value, instead of throwing a raw SyntaxError', () => {
+    const txData = { to: ROUTER, data: depositCalldata(), value: '0x' };
+    expect(() => assertEvmBridgeStepIntent(txData, intent)).toThrow(/malformed native value/);
+    try {
+      assertEvmBridgeStepIntent(txData, intent);
+    } catch (e) {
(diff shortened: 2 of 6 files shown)


gulshngill (approved):
**Reviewed + tested locally at `2522a13`. LGTM overall — no fail-open path found. A few in-scope items below.**

**End-to-end test (real funds, PR branch):** ran a fresh funded round trip through the new guard — 8 USDC Base → Hyperliquid (`approve` [0x962ec2a…](https://basescan.org/tx/0x962ec2acaf2c8581d8fdfa4177147751dc147507d19652884e7ec3648daf2b4c), `deposit` [0xe0339e3…](https://basescan.org/tx/0xe0339e3db8850b32e532fd260a937545f7ef2d883a903da5009a85d418325cf2)), then withdrew back. Verified on-chain: the approve signed **exactly 8,000,000 base units** (never unlimited), spender = the pinned router, and the **residual allowance is 0** after the pull. Return-leg balance reconciled to the base unit. Also replayed a tamper matrix against the real live quote (MAX approve, attacker spender, doubled approve/deposit pairs, redirected depositor, non-zero value, approve-only plan) — every one refused with the right error code.

**Constants provenance:** selector `0xe8017952` resolves to `depositErc20(address,address,uint256,bytes32)` (verified-contract entry in openchain/4byte; keccak recomputed) — matches the decoded layout exactly. Suggest updating the comments that call it `deposit(...)` to the real name, since the provenance is now publicly checkable. Router bytecode on Base dispatches this selector, and a fresh quote today matches all pins.

**Findings:**

1. **Quote/execute contract mismatch (medium-low).** `bridge quote` accepts any origin token on base→hyperliquid (`resolveBridgeToken` passes unknown symbols/addresses through, and the `--amount-unit usd` path even prices non-stables), but the new guard only ever passes USDC plans — so a non-USDC quote is mintable but permanently unexecutable. Suggest refusing non-USDC origin at quote time (mirroring the HL-origin lock) or documenting USDC-only in schema/usage text.
2. **Test wiring gap (minor).** Mutation check: deleting the `preflightEvmBridgeSteps` call site *or* the per-step `assertEvmBridgeStepIntent` in `processEvmStep` individually leaves the full suite green (only removing both together fails a test). The plan-shape bound (approve-only / amplification) is covered purely as a unit test, never through the execute command. One execute-path test feeding a tampered plan through the command mock would pin both layers.
3. **Uncoded approve-over-cap error (minor).** When the approve exceeds the anchor, the throw escapes from `encodeApproveCalldata` as a plain `Error` — `e.code` is `undefined` and the message lacks the "Request a new quote" action every sibling refusal has. Worth wrapping in a coded `CommandError`.
4. **Nits:** a non-null, non-numeric `requestedAmountBaseUnits` reaches `BigInt()` and throws a raw `SyntaxError` (fails closed, but inconsistent with the guard's own try/catch style); deposit calldata with dirty upper bits in the address words passes through unchanged (the approve branch re-encodes clean — deposits could normalize too).

None of these block the merge — the core guarantee ("at most the requested amount can leave the wallet") held under both adversarial unit testing and a real-money run.

PR #539, opened 2026-08-31
Enforce known-asset allowlist and per-payment cap for x402 auto-payment

kome12:
## Summary

x402 auto-payment previously signed EIP-3009 / Permit2 authorizations using the `amount`, `pay_to`, `network`, and `asset` fields taken verbatim from the server's 402 response, with no cap, allowlist, or confirmation. A malicious or compromised server (or a network-level injection in the `payment-required` header) could direct the CLI to silently sign a large transfer to an attacker-controlled address.

This PR adds a shared policy guard that evaluates every payment requirement before any key material is touched, applied across all three signing paths.

## What changed

**New module `src/x402-policy.js`** — the single source of truth for payment authorization:
- **Known-asset / known-network allowlist**: only signs payments whose `(network, asset)` pair appears in the existing token tables. Unknown assets or chains are refused with a clear message. This also provides trusted `decimals` for the USD conversion below.
- **Per-payment USD cap**: converts `amount` (base units) to USD using the trusted decimals and refuses if it exceeds a configurable ceiling. Default is `$1.00`; overridable via `NANSEN_X402_MAX_AMOUNT=<usd>` (or `=unlimited` to disable). Can also be persisted in wallet config as `x402MaxAmount`.
- **Optional recipient allowlist**: `NANSEN_X402_ALLOWED_PAYTO` (comma-separated, unset = no restriction).

**New module `src/x402-tokens.js`** — extracts the token tables into a leaf module to avoid circular imports, re-exported from `src/x402.js` for backward compatibility.

**Guard wired into all three signing paths** (`src/x402.js`, `src/privy.js`, `src/walletconnect-x402.js`). Each path logs the refusal reason to stderr (matching existing conventions) or throws on the WalletConnect path.

**New env vars documented** in `nansen wallet --help`.

## New env vars

| Variable | Default | Description |
|---|---|---|
| `NANSEN_X402_MAX_AMOUNT` | `1.00` | Max USD per auto-payment; `unlimited` to disable |
| `NANSEN_X402_ALLOWED_PAYTO` | unset | Comma-separated recipient allowlist (optional) |

## Tests

- `src/__tests__/x402-policy.test.js` — 20 unit tests covering the full policy surface: allowlist pass/fail, cap boundary (just over / just under), `unlimited` override, custom cap, payTo allowlist, 18-decimal token conversion, malformed amount, unknown network/asset.
- `src/__tests__/x402.test.js` (extended) — 2 integration tests: over-cap amount yields nothing and `createEvmPaymentPayload` is never called; within-cap amount yields a signature and the payload builder is called.
- `src/__tests__/privy.test.js` (updated) — existing "continues to next requirement on signing failure" test updated to use a known-allowlist network for the second requirement (was using a network not in the token table).

All 61 test files pass; lint clean.

## Possible follow-ups (out of scope here)

- Cumulative / session spend ledger — the per-payment cap does not stop repeated small payments across many 402 responses. A persisted running total with a daily ceiling is a larger, separate change.
- On-chain destination reputation checks — `NANSEN_X402_ALLOWED_PAYTO` is the pragmatic stand-in.
- Interactive confirmation (`NANSEN_X402_CONFIRM=1`) — step 3 from the design; intentionally deferred to keep this PR focused on the non-interactive security core.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


src/x402-tokens.js
```diff
@@ -0,0 +1,29 @@
...
+    { token: '0x8d0d000ee44948fc98c9b98a4fa4921476f08b0d', symbol: 'USD1', decimals: 18 }, // World Liberty Financial USD
+    { token: '0x55d398326f99059fF775485246999027B3197955', symbol: 'USDT', decimals: 18 }, // Tether USD
+    { token: '0x8AC76a51cc950d9822D68b83fE1Ad97B32Cd580d', symbol: 'USDC', decimals: 18 }, // Binance-Peg USD Coin
+  ],
+};
+
+// Known payment tokens on Solana.
+export const SVM_X402_TOKENS = {
+  solana: [
+    { token: 'EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v', symbol: 'USDC', decimals: 6 },
```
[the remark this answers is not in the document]
kome12: @AikidoSec ignore: This is a token address (USDC)

src/x402-policy.js
```diff
@@ -0,0 +1,135 @@
...
+ * with a human-readable, actionable reason when refused.
+ *
+ * Enforces in order: known (network, asset) allowlist, optional payTo
+ * allowlist, per-payment USD cap. Never signs; pure decision.
+ */
+export function evaluatePaymentRequirement(requirement) {
+  const network = requirement.network;
+  const asset = requirement.asset;
+  const payTo = requirement.payTo ?? requirement.pay_to;
+  const amountRaw = requirement.amount ?? requirement.maxAmountRequired;
```
Codier: **[P1] Cap bypass: the policy reads `amount ?? maxAmountRequired`, but the signers sign `amount || maxAmountRequired`**

The guard normalizes with `??`, so a falsy-but-present `amount` (`''` or `0`) is treated as $0.00, passes the cap, and returns `ok: true`. The signing paths then substitute `maxAmountRequired` for that falsy value and sign it instead:

- `walletconnect-x402.js` — `buildEIP712TypedData` and the `authorization.value` both use `requirement.amount || requirement.maxAmountRequired`
- `privy.js` — the `authorization.value` uses the same `||` expression

A malicious server sending `{ amount: "", maxAmountRequired: "1000000000000" }` ($1M in 6-decimal base units) therefore passes the policy at $0.00 while Privy/WalletConnect sign a $1,000,000 authorization. Reproduced on this branch:

```
amount="" -> policy.ok=true | typedData.message.value="1000000000000"
```

This reopens the exact vulnerability the PR exists to close. Fix: use `??` in the signers (`walletconnect-x402.js` typed data + authorization fields, `privy.js` authorization), or make the policy refuse falsy `amount` values outright. Note the local-EVM/SVM paths call `BigInt(requirements.amount)` directly, so `""` signs 0n there — inconsistent but not a bypass.

src/x402-policy.js
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Resolve the per-payment USD cap.
+ * Precedence: NANSEN_X402_MAX_AMOUNT env var > wallet config x402MaxAmount > default.
+ * Returns a positive number, or Infinity when explicitly set to "unlimited".
+ */
+export function resolveMaxAmountUsd() {
+  const env = process.env.NANSEN_X402_MAX_AMOUNT;
+  if (env !== undefined) {
+    if (env.trim().toLowerCase() === 'unlimited') return Infinity;
+    const n = Number(env);
```
Codier, line 63: **[P2] Empty `NANSEN_X402_MAX_AMOUNT` silently becomes a $0.00 cap**

`Number('')` is `0`, which satisfies `Number.isFinite(n) && n >= 0`, so `NANSEN_X402_MAX_AMOUNT=""` (or whitespace, e.g. an exported-but-unset variable in a script) yields cap `$0.00` and refuses every non-zero payment with `exceeds the $0.00 per-payment cap`. The comment above (`// fall through on garbage value`) claims garbage falls through to config/default, but the empty string never does.

Reproduced: with `NANSEN_X402_MAX_AMOUNT=""`, a $0.01 payment is refused with the `$0.00` message.

Fix: treat `env.trim() === ''` as unset before `Number(env)`, e.g. run the numeric check only on a non-empty string.

Codier (commented):
Review of the x402 payment-limits hardening: two actionable findings, no P0. One P1 (the per-payment USD cap is bypassable via amount/maxAmountRequired normalization split between the policy and the signers) and one P2 (empty NANSEN_X402_MAX_AMOUNT becomes a $0.00 cap). Both reproduced against this branch.

kome12:
@Codier Both findings from your review are fixed:

- **P1 (cap bypass)**: The `amount ?? maxAmountRequired` vs `amount || maxAmountRequired` mismatch between the policy guard and the signers is closed. Centralized amount resolution into `resolvePaymentAmount()` in `x402-policy.js`, used by the guard and every signing path (local EVM/SVM, Privy, WalletConnect) so they can no longer diverge.
- **P2 ($0 cap)**: `NANSEN_X402_MAX_AMOUNT=""` now correctly falls back to the default cap instead of `Number('') === 0`.

While hardening this further we also found and fixed the same class of bug for the `payTo`/`pay_to` field (added `resolvePayTo()`, same single-source-of-truth pattern) and a Solana-specific issue where addresses were being compared case-insensitively (base58 is case-sensitive, unlike EVM hex). Full details in the commit messages on the branch.

`nansen-pr-reviewer` came back clean on the latest commit. Would appreciate another look whenever you have time.

Codier (approved):
Approved after re-reviewing the four follow-up commits (d7b303a..8de0199).

**P1 — cap bypass via `??` vs `||` amount normalization: FIXED.**
`resolvePaymentAmount` / `resolvePayTo` in `x402-policy.js` are now the single source of truth, and every signing path consumes them (`buildEIP712TypedData`, Privy authorization, `createEvmPaymentPayload`, `createPermit2ExactPayload`, `buildUnsignedSvmTransaction`). A present-but-empty `amount` now falls through to `maxAmountRequired` in both the policy and the signers, so no signing path can sign a value the policy didn't cap. Reproduction confirms: `{ amount: "", maxAmountRequired: "1000000000000" }` is refused by the policy and would be signed identically by the signers.

**P2 — empty `NANSEN_X402_MAX_AMOUNT` => $0.00 cap: FIXED.**
Empty/whitespace env now falls through to wallet config / $1.00 default. Reproduction confirms `NANSEN_X402_MAX_AMOUNT=""` allows a $0.01 payment.

Also noted as correct: Solana mint / payTo comparisons are now case-sensitive (base58 is case-sensitive, unlike EVM hex) and the `isSvmNetworkString` duplication is deliberate to avoid an import cycle with `x402-svm.js`; I verified the import graph stays acyclic. Permit2 preflight now uses the resolved amount.

Validation: full test suite 62/62 files pass (2508 passed, 2 skipped); targeted x402 suite (policy, x402, evm, svm, walletconnect, privy) 102/102; eslint clean on changed files. One unrelated pre-existing flake (`bridge-quote.test.js`, network-dependent) failed once in an earlier run and passed in both subsequent full runs.

Out of scope (previously noted, would strengthen the guard but not blocking): per-request/session cumulative cap to bound many small payments across repeated 402s, and logging each signature / surfacing the low-balance check when the retry is rejected.

PR #569, opened 2026-09-03
fix: refuse re-executing an already broadcast swap quote

memosr:
## Problem

Swap quotes have no single-use guard. Bridge quotes do: `loadBridgeQuote` refuses a quote carrying `executedAt`, which `markBridgeQuoteExecuted` sets right after signing (added in #467, "preventing an accidental double-bridge on retry"). `loadQuote` in `trading.js` only checks the 1-hour TTL and the quote `type`.

The realistic trigger is the `RECEIPT_TIMEOUT` path added in #519. When the receipt poll times out:

- the transaction has already been broadcast (`executeTransaction` returned)

- `isFatalBroadcastError` re-throws past the per-quote retry loop and the command exits non-zero

- nothing writes back to the quote file, so it stays on disk and fully loadable until the TTL lapses

A retry with the same `--quote <id>` then re-signs under a freshly fetched nonce (`getEvmNonce` is called per execute, never pinned to the first run) and broadcasts a second, independently valid swap. This is not a byte-identical replay a node would drop.

Scope: EVM local wallet, Privy EVM, `--gasless` Relay, and WalletConnect. Jupiter/OKX-shaped Solana quotes are protected incidentally because the blockhash is baked into the quote, but the Relay Solana-bridge shape calls `fetchRecentBlockhash` live on each execute and has the same gap. `trade limit-order` is unaffected (no persisted quote-then-execute flow).

I could not determine whether the backend deduplicates on the `quoteId` sent in the execute request (#431 introduced it "for BI correlation", and `schema.json` documents "quotes are single-use" only for `perp bridge execute`, not for `trade execute`). Either way, `bridge.js` reaches the same backend and still guards client-side, so this reads as an oversight rather than a scoped decision.

## Fix

Mirrors the existing bridge pattern onto the swap quote file, which lives in the same `safeQuotesPath()` keyspace and is already keyed by quote id:

- `markQuoteExecuted(quoteId, progress)` sets `executedAt` in place on the same `{quoteId}.json` that `saveQuote` wrote. Best-effort: a write failure warns nothing and never crashes a run whose transaction is already out.

- `loadQuote` refuses a quote carrying `executedAt`, reporting when it was executed, any known tx hash, that the transaction may still be pending, and that a new quote is needed.

- Marking happens at broadcast time, not on success, so it covers the `RECEIPT_TIMEOUT` case. Two call sites: the WalletConnect wallet-auto-broadcast branch, and the shared `executeTransaction` call that every other path converges on.

Known residual gap, deliberately not addressed here: if `executeTransaction` throws after the backend has accepted the transaction, the quote is not marked. Narrowing that needs a pre-broadcast marker plus rollback and felt out of scope for a fix that mirrors an existing pattern. Happy to extend if you would prefer that.

## Tests

`src/__tests__/trading-quote-reuse.test.js` runs the EVM execute path twice against the same persisted quote with the RPC and Trading API mocked. Before the fix the second run produced a second broadcast; it now rejects with "already executed" and the broadcast count stays at 1. A second test covers the `RECEIPT_TIMEOUT` path specifically: the first run broadcasts and then times out, the second run is refused.

Test Files 63 passed (63)
Tests 2548 passed | 2 skipped (2550)

`npm run lint` clean. Changeset added (`patch`).

### DIFF
--- a/.changeset/fix-swap-quote-reuse.md
+++ b/.changeset/fix-swap-quote-reuse.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
+Refuse to re-execute a swap quote that has already been broadcast, mirroring the single-use guard `nansen bridge execute` already had. `nansen trade execute` now marks the quote as spent (`executedAt`) the instant a transaction is broadcast — before waiting for its receipt — so a `RECEIPT_TIMEOUT` (the tx is on-chain but the command exits non-zero) no longer leaves the quote replayable. Retrying the same `--quote <id>` after such a failure previously re-signed and re-broadcast the swap under a fresh nonce instead of being refused.
--- a/src/__tests__/trading-quote-reuse.test.js
+++ b/src/__tests__/trading-quote-reuse.test.js
@@ -1,0 +1,221 @@
+/**
+ * Regression test: `nansen trade execute --quote <id>` now has a single-use /
+ * idempotency guard on the persisted swap quote, mirroring `nansen bridge
+ * execute` (src/bridge.js's `executedAt` + `markBridgeQuoteExecuted`, see
+ * `loadBridgeQuote`). `loadQuote` (src/trading.js) checks `data.executedAt`
+ * and refuses an already-broadcast quote; `markQuoteExecuted` records the
+ * broadcast on the quote file the instant a transaction goes out — at
+ * broadcast time, not on success — so a post-broadcast failure (a REVERTED
+ * receipt, or a RECEIPT_TIMEOUT that aborts the process, see PR #519's
+ * isFatalBroadcastError) still leaves the quote marked spent.
+ *
+ * Without this guard, a user (or an agent retrying a failed command, which is
+ * the default behavior this CLI is designed for) who re-runs the exact same
+ * `--quote <id>` after seeing a RECEIPT_TIMEOUT error would re-sign and
+ * re-broadcast the swap. On EVM this isn't a harmless resend: the nonce is
+ * fetched fresh from the chain on every execute (`getEvmNonce`, 'pending'
+ * count), so a second run would get a genuinely new, independently valid
+ * nonce and produce a second, distinct, independently broadcastable
+ * transaction — not a byte-identical replay a node would reject.
+ *
+ * This test proves the fix by running the EVM execute path twice against the
+ * same saved quote id, with the RPC and Trading API mocked, and asserting the
+ * second call is refused before it ever reaches /execute.
+ */
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import fs from 'fs';
+import path from 'path';
+import os from 'os';
+import { saveQuote, buildTradingCommands, evmTxHash, getQuotesDir } from '../trading.js';
+import { createWallet, showWallet } from '../wallet.js';
+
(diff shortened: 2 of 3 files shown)


memosr:
Thanks, addressed in the latest commit.

On the non-Success branch: I could not prove either way from this repo whether a swap `Failed` response can carry a hash that was actually broadcast, since that depends on Trading API internals. But the shape is observed for approval broadcasts (`trading.test.js` mocks `{ status: 'Failed', txHash: '0xApprovalHash' }`), so it is not a shape the client invents. The two failure modes are asymmetric: marking a response that never went out costs a needless re-quote, while not marking one that did go out is a silent double broadcast. I went with the former, gated on `result.signature || result.txHash` being present.

While revisiting this I also documented a behaviour choice that was implicit before: a broadcast that later reverts on-chain still consumes the quote, because the nonce was burned. That means if every aggregator candidate reverts, the user must re-quote. Deliberate rather than an oversight, and the alternative (marking only on eventual return) would reopen the window this PR closes.

The wider residual gap in the PR description stands: if `executeTransaction` throws after the backend accepted the transaction, the quote is still not marked. Closing that needs a pre-broadcast marker plus rollback, which felt like a separate change. Happy to take it on if you would like it in scope.

kome12:
> Thanks, addressed in the latest commit.
> 
> On the non-Success branch: I could not prove either way from this repo whether a swap `Failed` response can carry a hash that was actually broadcast, since that depends on Trading API internals. But the shape is observed for approval broadcasts (`trading.test.js` mocks `{ status: 'Failed', txHash: '0xApprovalHash' }`), so it is not a shape the client invents. The two failure modes are asymmetric: marking a response that never went out costs a needless re-quote, while not marking one that did go out is a silent double broadcast. I went with the former, gated on `result.signature || result.txHash` being present.
> 
> While revisiting this I also documented a behaviour choice that was implicit before: a broadcast that later reverts on-chain still consumes the quote, because the nonce was burned. That means if every aggregator candidate reverts, the user must re-quote. Deliberate rather than an oversight, and the alternative (marking only on eventual return) would reopen the window this PR closes.
> 
> The wider residual gap in the PR description stands: if `executeTransaction` throws after the backend accepted the transaction, the quote is still not marked. Closing that needs a pre-broadcast marker plus rollback, which felt like a separate change. Happy to take it on if you would like it in scope.

Thanks for finding this, and for the thorough writeup. I reviewed it and can confirm that the guard closes it. Approving and merging.

On your three points:
- Marking the non-Success / Failed-with-hash branch: agreed, and the asymmetry is the right call. A needless re-quote is cheap, a silent second broadcast isn't. Gating on `result.signature || result.txHash` being present keeps it from consuming a quote that clearly never went out, which is a sensible bound.
- Revert still consumes the quote: also right. The nonce is burned either way, so re-signing the same quote would just race the reverted tx. Better to force a re-quote than reopen the window. Good call making it explicit in a comment rather than leaving it implicit.
- The residual post-broadcast-throw gap: let's keep this out of the current PR. Mirroring the existing bridge guard is exactly the right scope for this change, and the pre-broadcast-marker + rollback approach is a bigger, separate piece of work that touches both the swap and bridge paths. I've logged a follow-up on our side to pick it up. No need for you to expand scope here, though it's appreciated that you offered.

kome12 (approved):

PR #575, opened 2026-09-04
fix: fail closed on ambiguous broadcast failures in swap and bridge execute

kome12:
## Problem

The single-use quote guard marks a quote spent only *after* the broadcast call returns. If the broadcast instead **throws after the tx may already be live**, the quote is left unmarked — so a swap re-broadcasts the next candidate in-run, and a later `trade execute --quote <id>` (or an agent auto-retry) re-signs under a fresh nonce and broadcasts a **second, independently valid** transaction.

The concrete trigger: `/execute` retries a 502/503 and, if all attempts fail, throws `BROADCAST_FAILED`. A 502 on the **ack** (not the send) is indistinguishable at the client from "never sent" — but the tx is live. `bridge.js` has the same post-broadcast-throw shape on its EVM and signature legs.

## Fix — fail closed on ambiguity

The client can't distinguish "502, never sent" (safe to retry) from "502, sent but ack lost" (unsafe), so this is a policy choice, not detection. We fail closed: mark the quote spent and abort, trading a needless re-quote for never double-broadcasting. A **definitive** node rejection (a JSON-RPC error — the tx was refused, nothing is in flight) still leaves the quote reusable, so ordinary pre-broadcast errors don't burn a quote.

- **`trading.js`** — `BROADCAST_FAILED` is now a fatal broadcast error (aborts the candidate loop) and marks the quote spent in the catch (it throws out of `executeTransaction` before the normal mark runs).
- **`bridge.js`** — fail closed on non-definitive send failures across the EVM leg and both signature legs; a clean Hyperliquid rejection (`assertHyperliquidStepAccepted`) stays outside the guard, so it remains reusable.
- **`evmRpcCall`** — tags thrown errors with codes (`RPC_UNCONFIGURED` / `RPC_NETWORK_ERROR` / `RPC_HTTP_ERROR` / `RPC_JSON_ERROR`) so callers can tell a definitive rejection from an ambiguous gateway/transport failure.

## Testing

- New unit coverage: swap `BROADCAST_FAILED` fail-close (aborts without touching the next candidate, marks the quote, refuses re-execute); bridge ambiguous (non-JSON 502, network drop) → quote consumed, vs definitive (JSON-RPC rejection) → quote reusable.
- Full suite green (2579 passed / 2 skipped); lint clean.
- Live happy-path broadcast round-trips (Base ETH↔USDC, cross-chain ETH↔SOL) pass through the changed execute path.

Note: the ambiguous branch itself is only reachable during a real broadcast failure (impractical to force live without deliberately broadcasting), so it's verified via the unit tests that drive the exact failure shapes.

### DIFF
--- a/src/trading.js
+++ b/src/trading.js
@@ -205,11 +205,28 @@ export async function executeTransaction(params, { retries = 2, retryDelayMs = 1
       await new Promise(r => setTimeout(r, retryDelayMs));
     }
 
-    const res = await fetch(`${TRADING_API_URL}/execute`, {
-      method: 'POST',
-      headers,
-      body: JSON.stringify(params),
-    });
+    let res;
+    try {
+      res = await fetch(`${TRADING_API_URL}/execute`, {
+        method: 'POST',
+        headers,
+        body: JSON.stringify(params),
+      });
+    } catch (netErr) {
+      // The POST left this process but no response came back (a reset/timeout).
+      // That may have struck AFTER the backend received the signed tx and
+      // broadcast it — indistinguishable from "never sent" — so treat it as
+      // BROADCAST_FAILED, the same fail-closed class as a 502. Retrying re-sends
+      // the SAME signed bytes (a byte-identical replay a node dedupes), so a
+      // retry here can't itself double-broadcast; only exhausting them fails
+      // closed at the caller (isFatalBroadcastError → mark the quote spent).
+      lastError = Object.assign(
+        new Error(`Execute POST to /execute failed: ${netErr.message}`),
+        { code: 'BROADCAST_FAILED' }
+      );
+      if (attempt < retries) continue;
+      throw lastError;
+    }
 
     const text = await res.text();
     let body;
--- a/src/trading.js
+++ b/src/trading.js
@@ -263,12 +263,18 @@ export async function executeTransaction(params, { retries = 2, retryDelayMs = 1
     // to the !res.ok branch below, which would surface a nonfatal upstream code
     // and let the candidate loop broadcast the next quote on top of a live tx.
     if (res.status === 502 || res.status === 503) {
+      // Only append the simulation fee hint for a NON-JSON body. A structured
+      // JSON error (e.g. { code: "UPSTREAM_TIMEOUT" }) already explains itself
+      // via `details`; tacking "you may be out of SOL" onto a gateway timeout
+      // would misdirect the user.
       const chainType = params.chain && CHAIN_MAP[params.chain]?.type;
-      const feeHint = chainType === 'solana'
-        ? ' This often means the transaction failed simulation — check that you have enough SOL for fees (~0.005 SOL minimum).'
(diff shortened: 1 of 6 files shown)


src/trading.js
```diff
@@ -186,11 +205,28 @@ export async function executeTransaction(params, { retries = 2, retryDelayMs = 1
...
+      // closed at the caller (isFatalBroadcastError → mark the quote spent).
+      lastError = Object.assign(
+        new Error(`Execute POST to /execute failed: ${netErr.message}`),
+        { code: 'BROADCAST_FAILED' }
+      );
+      if (attempt < retries) continue;
+      throw lastError;
+    }
 
     const text = await res.text();
```
Codier: [P1] Can we include `res.text()` in the ambiguous-broadcast handling? `fetch()` can resolve after receiving the headers and then fail while reading a truncated/reset response body. I reproduced that here: the error has no code, so the candidate catch doesn't mark the quote or abort, and we can continue to the next signed transaction. That's the same double-broadcast window we're trying to close. I think body-read failures need to become `BROADCAST_FAILED` too, with a regression test.

Codier (changes requested):
I found two paths where we can still continue after an ambiguous broadcast result. I think these need to fail closed before we merge.

[P1] `src/trading.js:253`: I don't think JSON parseability makes a 502/503 definitive. A JSON 502 such as `{ code: "UPSTREAM_TIMEOUT" }` comes through this branch with that code, so the candidate loop treats it as nonfatal and tries the next quote. If the backend forwarded the signed transaction before timing out, we can double-broadcast. Can we classify 502/503 as `BROADCAST_FAILED` regardless of the response format and preserve the parsed body only as details?

src/trading.js
```diff
@@ -186,31 +205,92 @@ export async function executeTransaction(params, { retries = 2, retryDelayMs = 1
...
+      parsed = false;
+    }
+
+    // A 502/503 is ambiguous no matter the body SHAPE: the gateway may have
+    // forwarded the signed tx upstream before failing (a JSON 502 like
+    // { code: "UPSTREAM_TIMEOUT" } is exactly that case). Classify EVERY 502/503
+    // as BROADCAST_FAILED and keep the body only as details — never fall through
+    // to the !res.ok branch below, which would surface a nonfatal upstream code
+    // and let the candidate loop broadcast the next quote on top of a live tx.
+    if (res.status === 502 || res.status === 503) {
```
Codier: [P1] Can we fail closed for every 5xx response here, rather than only 502/503? A parseable 504 such as `{ code: "UPSTREAM_TIMEOUT" }` still falls through with that nonfatal code. I reproduced that on this head, and the candidate catch then tries the next quote even though a 504 specifically means the gateway may have forwarded the signed transaction and lost the ack. A structured 500 has the same ambiguity. I think all 5xx responses need to become `BROADCAST_FAILED`, with a regression test for 504.

Codier (approved):

PR #406, opened 2026-04-14
fix: wallet list routes human-readable summary to stderr, keeps only JSON on stdout

Nicolai1205:
Fixes #154

Rebased version of #289 (which had a merge conflict) onto current main.

## What
`nansen wallet list` was writing human-readable text to stdout alongside JSON, breaking agent pipelines that use `| jq .`.

## How
- Route human-readable wallet summary to `stderr` via `ttyOutput = console.error` dep
- Keep clean JSON output on `stdout` only
- Interactive users still see the summary (it goes to stderr which terminals display)
- Agent pipelines work: `nansen wallet list | jq .`

## Test
```
nansen wallet list        # human-readable summary on stderr, JSON on stdout
nansen wallet list | jq . # parses cleanly
```

All 929 tests pass.

PR #289, opened 2026-03-11
fix: wallet list routes human-readable summary to stderr, keeps only JSON on stdout

yodablocks:
Closes #154

## Problem
`nansen wallet list` was outputting human-readable text only (no JSON), 
breaking agent pipelines that rely on `| jq .`

## History
- PR #176 attempted to fix mixed stdout by removing human-readable output entirely — too aggressive
- PR #185 reverted it
- PR #176 commit explicitly noted a TODO: route human-readable text to stderr instead

## This PR completes that TODO

- Human-readable wallet summary → `stderr` (interactive users still see it)
- Clean JSON → `stdout` (agent pipelines work: `nansen wallet list | jq .`)

## Changes
- `src/wallet.js` — added `ttyOutput = console.error` dep, switched list handler to use it
- `src/__tests__/wallet.test.js` — added ttyOutput mock + explicit stderr/stdout separation test
- `.changeset/fix-wallet-list-stderr.md` — patch changeset

## Tests
929 tests pass, lint clean.

PR #576, opened 2026-09-04
fix: route wallet list summary to stderr, emit JSON on stdout

erhnysr:
## The bug

`nansen wallet list` writes its human-readable summary (wallet names, EVM/Solana addresses, the `★` default marker) to **stdout** via `log()`, and returns `undefined` so no structured JSON is emitted. That mixes prose and data on the same stream, so agent pipelines that expect parseable output break:

```console
$ nansen wallet list | jq .
parse error: Invalid numeric literal at line 1, column 5
```

Every other data-returning command emits clean JSON on stdout; `wallet list` is the odd one out, which makes it unusable in `| jq`-style automation.

## Prior history — why this keeps coming back

This was reported before in #289 and #406, and both were closed on the belief that the fix had already landed via #397. That belief is incorrect, and I verified it two ways before writing this:

- **#397's actual diff** only extracted a `requireWalletFile()` helper for the file-existence checks in `showWallet`, `exportWallet`, `setDefaultWallet`, and `deleteWallet` (plus a `deriveEvmAddress()` helper in `transfer.js`). It is a self-described "no behavior changes" refactor and **never touches `listWallets` or the `list` handler**.
- A full `git log -L :listWallets:src/wallet.js` trace shows `listWallets` was only ever touched by two commits: its original creation, and one Privy PR that added a `provider` field. **No commit in its history ever introduced stderr/stdout separation.**

So this is a **fresh fix**, not a rebase or reopen of the older PRs — the output path they were meant to fix was never actually modified.

## The fix

In the `list` handler (`src/wallet.js`):
- Route the human-readable summary to **stderr** via `process.stderr.write` (the file's existing convention).
- Return a structured `{ wallets }` value so `runCLI`'s normal success path renders clean JSON to **stdout** — mirroring how research/data commands emit their output.
- The empty case prints its `No wallets found` hint to stderr and still returns `{ wallets: [] }`, so **stdout is always valid JSON**.

After the fix:
```console
$ nansen wallet list | jq '.data.wallets[].name'
"main"
```
(the summary still renders for humans — it's just on stderr now.)

## Testing & verification

- **Failing-first test** added driving the real `runCLI`: before the fix it fails with `SyntaxError: Unexpected end of JSON input` (the summary went to real stdout, the JSON sink was empty) — reproducing the exact bug.
- After the fix, the wallet-list tests pass (**4 passing**), including an updated Privy test that now asserts the provider tag lands on stderr.
- **Full suite: 2623 passed, 2 skipped, 0 failed** (64 files).
- **`npm run lint`: clean.**
- Changeset added (`patch`), user-facing.

### DIFF
--- a/.changeset/fix-wallet-list-stdout-json.md
+++ b/.changeset/fix-wallet-list-stdout-json.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
+Fix `nansen wallet list` polluting stdout with its human-readable summary, which made the output impossible for an agent to `JSON.parse`. The decorated summary (wallet names, EVM/Solana addresses, default marker) now goes to stderr, and the command returns a structured `{ wallets }` value so stdout carries only clean JSON — matching how research commands emit their data. The empty case (`No wallets found`) likewise prints its hint to stderr and still emits `{ "wallets": [] }` on stdout.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,7 +24,7 @@ Entry point is `src/index.js`.
 - **ESM only** — `import`/`export`, no TypeScript, no transpilation
 - **BigInt for token amounts** — never floating point
 - **Research commands** — return data objects, CLI layer formats via `formatOutput()` to stdout
-- **Operational commands** (trade, wallet, login) — print human-readable text via `log()` to stdout, return `undefined`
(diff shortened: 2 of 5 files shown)


erhnysr:
Prior attempts at this same fix, both closed on the mistaken belief it had landed via #397:

- #289
- #406

I left a note on each explaining why #397 did not actually fix it (it only refactored the existence-check helper and never touched `listWallets`), and pointing here. Cross-linking so the full history is visible from this PR.

erhnysr:
Bot auto-approved this a couple weeks ago (Sep 4) — just flagging in case it needs a human merge decision. Also cross-linked from the two prior attempts (#289, #406) that were closed on the mistaken belief this was already fixed via #397, so the full history is visible here. Happy to rebase if main has moved.

PR #579, opened 2026-09-04
fix: alerts numeric flags silently produce wrong results on invalid input

### LINKED ISSUE #578: alerts list --limit/--offset and numeric range flags silently produce wrong results on invalid input
## Summary

`nansen alerts list --limit`/`--offset` and 9 numeric flags on `nansen alerts create`/`update` accept non-numeric input without validation, producing silently wrong results instead of an error.

## Reproduction

1. `nansen alerts list --limit abc`
   Expected: an error like `--limit must be a positive integer`.
   Actual: an empty list -- indistinguishable from genuinely having zero alerts. `Number('abc')` is `NaN`, and `Array.prototype.slice(0, NaN)` evaluates the end index as `0`.

2. `nansen alerts list --offset -2`
   Expected: an error rejecting the negative offset.
   Actual: silently returns the *last* 2 alerts. `Array.prototype.slice()` treats a negative start index as "count from the end".

3. `nansen alerts create ... --market-cap-min notanumber`
   Expected: an error like `--market-cap-min must be a number`.
   Actual: the alert is created successfully, with the market-cap-min threshold silently disabled. `Number('notanumber')` is `NaN`, which passes the existing `!= null` "is this set" check (NaN is neither null nor undefined) -- but `JSON.stringify({min: NaN})` serializes to `{"min":null}`, so the value sent over the wire is `null`, not the number the user typed and not an error either.

The same pattern (case 3) affects all of: `--market-cap-min/max`, `--usd-min/max`, `--token-amount-min/max`, `--fdv-min/max`, `--inflow-*-min/max`, `--outflow-*-min/max`, `--netflow-*-min/max`, `--token-age-min/max` (7 range-flag call sites plus 2 standalone token-age blocks in `src/commands/alerts.js`).

## Root cause

`src/commands/alerts.js` calls `Number(optionValue)` in several places (the shared `buildRange()` helper, two standalone `token-age` blocks, and the `list` handler's pagination) without checking whether the result is actually a valid number, unlike the equivalent pagination helpers in `src/query-options.js` and `src/commands/research.js`, which both validate with `Number.isInteger()` before using the value and throw a `NansenError` otherwise.

## Suggested fix

Validate with `Number.isFinite()` (range flags, which can be decimals) / `Number.isInteger()` (`--limit`/`--offset`, which must be whole numbers) and throw a `NansenError` on failure, matching the existing `buildPagination` idiom already used elsewhere in the CLI. I have a fix ready and will open a PR referencing this issue.

teyrebaz33:
## Summary

Fixes #578.

`src/commands/alerts.js` called `Number(optionValue)` in several places without checking whether the result was actually a valid number: the shared `buildRange()` helper (used by 7 numeric range flags on `alerts create`/`update`), two standalone `--token-age-min/max` blocks, and the `list` handler's `--limit`/`--offset` pagination.

A non-numeric value produced `NaN`, which is neither `null` nor `undefined`, so it silently passed the existing "is this flag set" checks. For the range flags, `JSON.stringify({min: NaN})` serializes to `{"min":null}`, so the alert was created with that threshold silently disabled instead of erroring. For `--limit`, `Array.prototype.slice(0, NaN)` evaluates the end index as `0`, so a typo'd `--limit` silently returned an empty list. For `--offset`, a negative value made `Array.prototype.slice()` count from the end of the array, so `--offset -2` silently returned the *last* 2 alerts.

## Fix

Validate with `Number.isFinite()` (range flags, which can be decimals) / `Number.isInteger()` (`--limit`/`--offset`, which must be whole numbers) and throw a `NansenError` on failure -- matching the existing `buildPagination` idiom already used in `src/query-options.js` and `src/commands/research.js`. I checked for the same unvalidated-`Number()` pattern elsewhere in `src/commands/` and confirmed `research.js`'s pagination already validates correctly, so this PR's scope (alerts.js only) is complete.

## Test plan

- Added regression tests covering all 9 fixed call sites (non-numeric range flags, non-numeric/negative token-age, non-numeric/negative/zero/non-integer `--limit` and `--offset`), plus a test confirming numeric strings (how the real CLI parser passes flag values) still work.
- `npx eslint src/commands/alerts.js src/__tests__/cli.internal.test.js` -- clean.
- `npx vitest run src/__tests__/cli.internal.test.js` -- 484/484 passed.
- Full suite (`npm test`): 2639 passed, 2 skipped, 0 failed.
- `npm install` produced only unrelated `package-lock.json` metadata noise, reverted before committing; `package.json` itself was not touched.

https://claude.ai/code/session_0141eH4NcQy4GXgusCGbG1pm

### DIFF
--- a/.changeset/fix-alerts-numeric-validation.md
+++ b/.changeset/fix-alerts-numeric-validation.md
@@ -1,0 +1,9 @@
+---
+"nansen-cli": patch
+---
+
+Fix `nansen alerts list --limit`/`--offset` and the numeric range flags on `nansen alerts create`/`update` (`--market-cap-min/max`, `--usd-min/max`, `--token-amount-min/max`, `--fdv-min/max`, `--inflow/outflow/netflow-*-min/max`, `--token-age-min/max`) silently accepting non-numeric input instead of erroring.
+
+`--limit abc` returned an empty alert list (`Array.prototype.slice(0, NaN)` evaluates the end index as `0`), indistinguishable from genuinely having zero alerts. `--offset -2` silently returned the *last* 2 alerts (a negative `slice()` start counts from the end of the array) instead of erroring on an invalid pagination offset. A non-numeric range flag like `--market-cap-min notanumber` passed validation (`NaN != null` looks "set") but serialized to `{"min":null}` over JSON, silently creating the alert with that threshold disabled rather than the value the user typed.
+
+All of these now throw a clear `--<flag> must be a number` / `must be a positive integer` / `must be a non-negative integer` error instead of silently producing a wrong result, matching the validation style already used by `--limit`/`--page` elsewhere in the CLI (`src/query-options.js`, `src/commands/research.js`).
--- a/src/__tests__/cli.internal.test.js
+++ b/src/__tests__/cli.internal.test.js
@@ -305,6 +305,26 @@ describe('buildSmTokenFlowsData', () => {
     const result = buildSmTokenFlowsData({});
     expect(result).toEqual({});
   });
+
+  it('rejects a non-numeric range flag instead of silently producing NaN (regression)', () => {
+    // Before the fix: buildRange did `Number(minVal)` with no check. NaN
+    // passes isRangeSet()'s `!= null` test (NaN is neither null nor
+    // undefined), so a typo'd threshold looked "set" all the way through
+    // validation -- but JSON.stringify({min: NaN}) serializes to
+    // {"min":null}, silently disabling the threshold the user asked for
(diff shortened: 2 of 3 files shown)


kome12:
Thank you for this, @teyrebaz3, we really appreciate you reporting and fixing it, and the write-up was spot on.

This overlapped with a couple of other community PRs, so we're consolidating to keep the history clean. Closing this as superseded by the changes we're merging. Your PR helped confirm the scope of the fix, and it's much appreciated.

We'd love to have your next contribution 🙏

PR #580, opened 2026-09-04
feat: add generated bash/zsh/fish shell completions

gulshngill:
Adds `nansen completion <bash|zsh|fish>`, closing the gap against `gh completion`, Railway, AWS CLI, and Terraform.

Refs **API-273**.

## What

```bash
nansen completion bash     # prints the script to stdout
nansen completion zsh
nansen completion fish
nansen completion          # usage + install instructions for all three
```

The scripts are **generated from `src/schema.json`** — the same source of truth `nansen schema` and `--help` read — so they cannot drift from the documented command surface. They cover:

- nested subcommands, to full depth (`nansen research token screener`, `nansen trade limit-order create`)
- per-command flags and global flags (`--pretty`, `--fields`, `--limit`, …)
- enum values a flag accepts (`--tif Gtc|Ioc|Alo`, `--side buy|long|sell|short`, `--chain …`)

Each script walks the command line to resolve the current command path, skipping option values so they are never mistaken for a subcommand — `nansen --pretty research smart-money <TAB>` and `nansen research --fields screener <TAB>` both resolve correctly.

zsh and fish also carry per-item descriptions from the schema; bash emits names only (its default menu has nowhere to show them).

`completion` is treated as an offline command: no network, no disk writes, and the background update check and telemetry are skipped so nothing else can land on a stdout that gets piped into a shell.

## Install

```bash
# bash
echo 'eval "$(nansen completion bash)"' >> ~/.bashrc
# or: nansen completion bash > /etc/bash_completion.d/nansen

# zsh
nansen completion zsh > "${fpath[1]}/_nansen" && compinit

# fish
nansen completion fish > ~/.config/fish/completions/nansen.fish
```

## Testing in CI

`src/__tests__/completion.test.js` (35 tests) keeps generation honest without a human tabbing through a terminal:

1. **Syntax** — `bash -n` on the generated script every run; `zsh -n` and `fish --no-execute` when those shells exist on the runner, plus a shell-agnostic block-balance check for fish so the fish path is never wholly unchecked.
2. **Behaviour** — the bash script is sourced in a subprocess and driven the way bash drives it (`COMP_WORDS`/`COMP_CWORD` → `COMPREPLY`), asserting top-level commands, prefix matching, nested subcommands, per-command options, enum values, and the two path-walking edge cases above.
3. **Drift** — every top-level command the CLI actually dispatches must appear in the completion tree or in an explicit exclusion list, and every flag `parseArgs` treats as valueless must be valueless to the walker. Adding a command without updating the schema fails the suite.
4. **Quoting** — `schema.json` descriptions contain `'`, `"`, `` ` ``, `$` and `:`; a test pins that a quote-bearing description survives into zsh and fish source escaped.

## Assumptions (issue left these open)

- **Deprecated top-level aliases are not completed.** `nansen token …`, `nansen quote …` etc. still work, but completion is a discoverability surface and suggesting them teaches the spelling we're retiring. `bridge-status` is excluded on the same grounds — it's an undocumented top-level alias of `trade bridge-status`. The list is explicit in `EXCLUDED_COMMANDS`.
- **Six commands the CLI dispatches aren't in `schema.json`** (`login`, `logout`, `schema`, `cache`, `changelog`, `help`). They're declared in `UNSCHEMA_COMMANDS` in the generator rather than added to `schema.json`, because the schema help path runs before their hand-written `--help` text and adding them would change existing output. A test fails if either list drifts.
- **`--chain` under `research` falls back to `schema.chains`.** Research `--chain` options declare no `enum`, and `schema.chains` is exactly the list those endpoints accept. Trade/bridge chains are narrower, so they get only their own declared enums rather than a guess.
- **Enum completion is only as good as `schema.json`.** Options that document a closed set in prose but declare no `enum` (e.g. `trade limit-order create --trigger-condition`, documented as `above`/`below`) complete nothing. Adding `enum` to those entries is a follow-up that improves `--help` and completion together.

## Other changes

- `parseArgs`' 20-clause valueless-flag condition is now an exported `VALUELESS_FLAGS` set — the generator's walker needs the same list, and the test asserts they stay in step.
- `schema.json`, `HELP`, and the README gain the `completion` command.
- Changeset: `minor`.

## Checks

```
npm run lint     # clean
npm test
 Test Files  65 passed (65)
      Tests  2670 passed | 3 skipped (2673)
```

(No build step — ESM, no transpilation. `npm pack --dry-run` confirms `src/commands/completion.js` ships.)

### DIFF
--- a/.changeset/generated-shell-completions.md
+++ b/.changeset/generated-shell-completions.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": minor
+---
+
+Add `nansen completion <bash|zsh|fish>`, which prints a shell completion script generated from the CLI's own command schema. Completions cover nested subcommands, per-command flags, global flags, and the enum values a flag accepts.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,10 +51,12 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      # zsh ships on ubuntu-latest; fish does not. Without it the fish
-      # completion tests skip and the fish script goes unexercised.
-      - name: Install fish for completion tests
-        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends fish
+      # The zsh and fish completion tests skip when the shell is missing, so
+      # install both and fail here if either is absent rather than skip quietly.
+      - name: Install zsh and fish for completion tests
+        run: |
+          sudo apt-get update && sudo apt-get install -y --no-install-recommends fish zsh
+          zsh --version && fish --version
 
       - name: Run tests
         run: npm test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,11 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      # zsh ships on ubuntu-latest; fish does not. Without it the fish
+      # completion tests skip and the fish script goes unexercised.
+      - name: Install fish for completion tests
+        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends fish
+
       - name: Run tests
         run: npm test
 
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ echo 'eval "$(nansen completion bash)"' >> ~/.bashrc
 
 # zsh
 nansen completion zsh > "${fpath[1]}/_nansen" && compinit
-# or: echo 'eval "$(nansen completion zsh)"' >> ~/.zshrc
+# or, in ~/.zshrc after the compinit line: eval "$(nansen completion zsh)"
 
 # fish
 nansen completion fish > ~/.config/fish/completions/nansen.fish
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ nansen trade quote --chain solana --from SOL --to USDC --amount 1000000000
 nansen trade execute --quote <quoteId>
 nansen wallet <subcommand> [options]
 nansen mcp install <client>           # add the Nansen MCP server to Claude Code/Desktop or Cursor
+nansen completion <bash|zsh|fish>     # shell completions (no API key needed)
 nansen schema [command] [--pretty]    # full command reference (no API key needed)
 ```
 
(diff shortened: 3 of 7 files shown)


gulshngill:
## Local test pass: what was run, what broke, what was fixed

Checked the branch out in a worktree and drove the generated scripts through real shells rather than relying on the unit tests alone.

### Coverage

| Check | How | Result |
|---|---|---|
| `npm test`, `npm run lint` | worktree, node 24 | 65 files / 2681 tests pass, lint clean |
| CLI paths | `nansen completion`, `completion --help`, `completion bash --help`, `completion tcsh`, `schema completion`, first-run with empty `HOME` and no API key | usage, schema help, `INVALID_PARAMS` exit 1, clean stdout (no update-check noise) |
| bash 5.3 (Homebrew) and bash 3.2 (macOS stock) | `bash -n`, then `_nansen_complete` driven via `COMP_WORDS`/`COMP_CWORD`, 14 cases each, plus `eval "$(nansen completion bash)"` install | all correct on both versions |
| zsh 5.9 | `zsh -n`, then real compsys tab completion in a `zpty` pty, 12 cases, both the `fpath` install and the `eval` install | all correct |
| fish 4.9 | `fish --no-execute`, then `complete -C` for 13 cases, plus autoload from `~/.config/fish/completions/nansen.fish` | **4 wrong, stderr noise on every tab** (see below) |

Cases covered nested subcommands, per-command and global options, enum values (`--tif`, `--side`, research `--chain`), prefix matching, and the path walker across `--pretty`, `-p`, `--fields <value>`, `--chain <value>`.

### Findings and fixes

**1. fish: path walker broken whenever a flag precedes the cursor (fixed in 19131ac).**
Both walkers called `string split ' ' (__nansen_flags)`. The flag list starts with `--all`, so fish parsed the list itself as options, printed `string split: ... unknown option` to stderr on every completion, and left the list empty. Every `-*` token then swallowed the next word:

- `nansen --pretty research smart-money <TAB>` offered top-level commands instead of `netflow`, `holdings`, …
- `nansen -p research <TAB>` same
- `nansen research token --pretty <TAB>` returned nothing instead of the token subcommands

Fix: a `__nansen_is_flag` helper that passes `--` to `string split`. Re-ran all 13 fish cases after the fix: all correct, stderr empty.

**2. zsh: `local path` shadowed `PATH` (fixed in 19131ac).**
zsh ties the array `path` to `PATH`, so `PATH` was empty inside `_nansen` for the duration of every completion. It worked only because compsys uses builtins; any external command reached during completion would have failed with `command not found`. Renamed to `cmdpath`. The new zsh behaviour test asserts `PATH` is intact inside the completion function and fails against the previous script.

**3. Single-dash tokens (fixed in 19131ac).**
`parseArgs` treats every `-x` token as valueless, but the walkers only knew five short aliases, so `nansen -x research <TAB>` dropped the path. Walkers now treat any single-dash token as valueless, matching the parser exactly.

**4. Docs: the zsh `eval` install must run after `compinit`** (fixed in 19131ac). Otherwise `compdef: command not found`. README, `nansen completion` usage, and the script header now say so.

**5. Dead `--help` branch** (fixed in 19131ac). `runCLI` answers `--help` from `schema.json` before the command runs, so the handler's `flags.help` branch and its test were unreachable. Removed.

**6. CI never exercised zsh or fish** (fixed in 19131ac and 9f8afbb). `ubuntu-latest` ships neither shell, so the zsh and fish tests were skipping on every run. The test job now installs both and prints their versions so a missing shell fails the job instead of skipping quietly. The pr-reviewer bot's high finding is the project's mandatory block on any `.github/workflows/` change; it notes no code change is needed.

### Tests added

`completion.test.js` went from 36 to 46 tests: zsh behaviour (5, driven outside compsys with a stubbed `_describe`, including the `PATH` assertion), fish behaviour (5, via `complete -C`, asserting empty stderr), and the single-dash bash case.

### Results after the fixes

- Local: `npm test` 65 files / 2681 passed, `npm run lint` clean, `completion.test.js` 46/46 with zsh and fish present.
- bash 5.3 and 3.2, zsh 5.9 (real compsys), fish 4.9: every case above correct, fish stderr empty.
- CI on 9f8afbb: lint and test (node 20, 22, 24) green. `completion.test.js` ran 46/46 with 0 skipped on all three jobs, against zsh 5.9 and fish 3.7.0 on the runner, so fish 3 is covered in CI and fish 4 locally. The pr-reviewer check shows as failed only for the procedural workflow-file block described in item 6.

Follow-ups, not blocking: 140 options and `research points` have no description in `schema.json` (zsh and fish list them bare), and the `mcp install` client names are prose-only in the schema so they don't complete.

PR #581, opened 2026-09-04
feat(schema): describe every option, complete mcp client positionals

gulshngill:
Follow-up to #580 (merged), which surfaced two gaps in `src/schema.json`: 140 options had no description (zsh and fish list them bare), and the `mcp install` client names were prose-only so nothing could complete them.

## What

**Descriptions and types for every option that lacked one (140).** All core research options (`--address`, `--chain`, `--days`, `--token`, `--market-id`, `--timeframe`, the prediction-market screener filters, …) and `wallet send --to/--chain/--amount` were declared as bare `{ "required": true }`, `{ "default": … }`, or `{}`. Each now carries `type` and a one-line `description` written from the handler code (what the value is, units, ranges). `research points` gets a group description. `nansen schema` no longer has undescribed parameters, and zsh/fish completions show a description for every flag.

**Accepted values declared where the code enumerates them.** `wallet send --chain` (`evm|solana|ethereum|base`), `research search --type` (`any|token|entity`), and the prediction-market `--neg-risk` filter (`true|false`) now declare `enum`, so `--help` prints them and completion offers them. `--neg-risk` is typed `boolean` like the other `resolveBooleanOption` flags; the completion generator treats a boolean that also declares an enum as value-taking, so `--neg-risk <TAB>` still offers `true`/`false` while plain booleans such as `--premium-labels` stay valueless.

**Positional completion.** `mcp install` and `mcp uninstall` declare their client as `args: [{ name, required, enum, description }]`. The completion generator reads `args` enums and offers them like subcommands in bash, zsh, and fish, without extending the command path, so `nansen mcp install cursor --<TAB>` still offers `--dry-run`. `--help` output is unchanged (the schema help renderer ignores `args` for now).

**Two corrections found while writing the descriptions:**
- `research token ohlcv --timeframe` declared default `1h`; the CLI and API default to `1d`. Schema now says `1d`.
- `--neg-risk true` was inverted. `parseArgs` JSON-parses `true` into a boolean, and the handler compared it with `=== 'true'`, so `true` sent `neg_risk: false`. Now resolved through the existing `resolveBooleanOption` helper (same as `--premium-labels`), so `--neg-risk`, `--neg-risk true|false`, and `--neg-risk 1|0` all work, and covered by tests. Worth fixing here because this PR is what starts advertising the flag's values.

## How the schema was edited

Line-based, not a JSON rewrite: `JSON.stringify` would have exploded every inline `enum` array into multi-line form and drowned the diff. Only the targeted option blocks changed; a structural comparison against `HEAD` confirms the only altered existing value is the ohlcv default. Key order matches the file's convention (`type`, `default`/`required`/`enum`, `description`).

## Verification

- `npm test` and `npm run lint` clean; `completion.test.js` 50 tests (4 new: positional values in the spec and in each shell, and that the path survives a positional). New `prediction-market-neg-risk.test.js` (6 tests) fails on the previous `cli.js` for the `true` cases and passes with the fix.
- Drove the regenerated scripts in bash 5.3 and macOS bash 3.2, zsh 5.9 (real compsys), and fish 4.9: `mcp install <TAB>` offers the three clients, `mcp install cursor --<TAB>` offers `--dry-run`, `wallet send --chain <TAB>`, `research search --type <TAB>`, and `--neg-risk <TAB>` offer their enums.
- `nansen research token ohlcv --help` now shows `--timeframe (1d)`; `nansen wallet send --help` shows `--chain* [evm|solana|ethereum|base]`.

## Not in this PR

- `generateSubcommandHelp` (used for `nansen research <cat> <sub> --help`) prints `--name (default) [enum]` without descriptions, so the new text reaches `nansen schema` and completions but not that help view. Rendering it there is a small follow-up.
- `wallet send` also accepts `--token`, `--wallet`, `--max`, and `--dry-run` per its hand-written help; the schema still omits them.

### DIFF
--- a/.changeset/schema-descriptions-and-positional-completion.md
+++ b/.changeset/schema-descriptions-and-positional-completion.md
@@ -1,5 +1,5 @@
 ---
-"nansen-cli": patch
+"nansen-cli": minor
 ---
 
 Every command option in `nansen schema` now carries a type and a description (140 research and `wallet send` options had neither), and the `research points` group is described. `research token ohlcv --timeframe` documents its real default (`1d`), and `wallet send --chain`, `research search --type`, and the prediction-market `--neg-risk` filters declare the values they accept, so `--help` and shell completion offer them. Fixed `--neg-risk true` on the prediction-market screeners, which was sent to the API as `neg_risk: false` because the parsed boolean was compared against the string `'true'`. `nansen mcp install` and `nansen mcp uninstall` declare their positional client in the schema, and `nansen completion` scripts now complete it (`claude-code`, `claude-desktop`, `cursor`) in bash, zsh, and fish.
--- a/.changeset/schema-descriptions-and-positional-completion.md
+++ b/.changeset/schema-descriptions-and-positional-completion.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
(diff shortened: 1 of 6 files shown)


gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
@nansen-pr-reviewer re-review

PR #597, opened 2026-09-07
fix(mcp): guard saved API key from disclosure to custom verify URLs

kome12:
## Problem

`nansen mcp verify --url <custom-url>` forwarded a **saved** API key (from `nansen login` / `NANSEN_API_KEY` / config file) to any user-supplied MCP endpoint in the authenticated `tools/call` request's `NANSEN-API-KEY` header — with only a warning, and over plain HTTP. A user (or automation) induced to verify a malicious URL would disclose their key to whoever controls that host, who could then make Nansen API calls billed to that user.

A saved key is trusted for the default Nansen endpoint only; supplying a custom URL is not itself authorization to send that credential elsewhere.

## Fix

Added a credential-disclosure guard before the authenticated `tools/call`:

- **Consent gate** — a saved key is withheld from a non-default `--url` unless `--send-api-key` is passed. This covers loopback too, so there is no silent drive-by even to `localhost`.
- **HTTPS hard-block** — no key is ever sent over plain `http://` to a non-loopback host, even with `--send-api-key`. Plain HTTP is permitted only for loopback dev endpoints (`localhost`, `127.0.0.0/8`, `::1`).
- An inline `--api-key` is unaffected — supplying it explicitly is itself consent, so the fresh-key workflow is unregressed.

Withheld cases emit an `error` check, skip the authenticated call, and exit non-zero with actionable guidance naming the destination origin and how to authorize it.

## Changes

- `src/mcp-verify.js` — destination classification helpers, `sendApiKey` param, and the two gates.
- `src/commands/mcp.js` — `--send-api-key` flag wired through, usage text updated.
- `src/cli.js` — registered `send-api-key` as a valueless flag.
- `src/schema.json` — documented the flag and added an example.
- `src/__tests__/mcp-verify.test.js` — 7 regression tests (withhold without flag, withhold on loopback, send with flag, loopback-http with flag, HTTP-to-public hard block, inline key still sends, CLI flag threading).

## Testing

- `npm test` — all passing (added 7 tests).
- `npm run lint` — clean.

### DIFF
--- a/src/commands/mcp.js
+++ b/src/commands/mcp.js
@@ -219,9 +219,12 @@ export function buildMcpCommands(deps = {}) {
     // `false` as a positional arg while the flag still reads as present, which
     // would authorize the very disclosure the caller meant to decline. Reject
     // any positional args so that footgun fails loudly instead of leaking.
+    // Never echo the argument: a plausible misuse is `--send-api-key "$KEY"`,
+    // which lands the real key in extraArgs[0]; interpolating it would leak the
+    // credential to stdout (and to logs under --json).
     if (extraArgs.length > 0) {
       throw new CommandError(
-        `Unexpected argument: ${extraArgs[0]}. \`nansen mcp verify\` takes no positional arguments — --send-api-key is a valueless flag, do not pass it a value. Usage: nansen mcp verify [--api-key <key>] [--url <url>] [--send-api-key] [--json]`,
+        '`nansen mcp verify` takes no positional arguments — --send-api-key is a valueless flag, do not pass it a value. Usage: nansen mcp verify [--api-key <key>] [--url <url>] [--send-api-key] [--json]',
         'INVALID_PARAMS',
       );
     }
--- a/src/commands/mcp.js
+++ b/src/commands/mcp.js
@@ -36,7 +36,7 @@ const MCP_USAGE = `nansen mcp — Install the Nansen MCP server into a local MCP
 USAGE:
   nansen mcp install <client>     Add the Nansen MCP server to the client's config
   nansen mcp uninstall <client>   Remove the Nansen MCP server from the client's config
-  nansen mcp verify [--api-key <key>] [--url <url>] [--json]
+  nansen mcp verify [--api-key <key>] [--url <url>] [--send-api-key] [--json]
                                   Verify the hosted MCP server and API key
 
 CLIENTS:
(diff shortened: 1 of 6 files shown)


kome12:
Thanks for the review — both addressed in 0222586.

**Medium (loopback regex):** Tightened to only accept valid octets (0–255). Worth noting: an invalid-octet host like `127.999.999.999` is actually rejected by `new URL()` before it ever reaches the classifier, so the loose pattern wasn't reachable in practice — but the tighter regex is clearer intent, and I went further: the guard now **fails safe** for any custom URL that isn't provably secure (unparseable, or a non-https / non-loopback-http scheme), rather than relying on `fetch` to reject it later. Added a regression test for that path.

**Low (semver bump):** Agreed — bumped the changeset to `minor`, since this adds the user-facing `--send-api-key` flag and changes observable `verify` behavior.

src/commands/mcp.js
```diff
@@ -212,7 +214,17 @@ export function buildMcpCommands(deps = {}) {
...
 
-  const verify = async (flags, options) => {
+  const verify = async (flags, options, extraArgs = []) => {
+    // --send-api-key is a valueless flag: `--send-api-key false` parses the
+    // `false` as a positional arg while the flag still reads as present, which
+    // would authorize the very disclosure the caller meant to decline. Reject
+    // any positional args so that footgun fails loudly instead of leaking.
+    if (extraArgs.length > 0) {
+      throw new CommandError(
+        `Unexpected argument: ${extraArgs[0]}. \`nansen mcp verify\` takes no positional arguments — --send-api-key is a valueless flag, do not pass it a value. Usage: nansen mcp verify [--api-key <key>] [--url <url>] [--send-api-key] [--json]`,
```
Codier: `--send-api-key` is presence-only, so a plausible invocation is `--send-api-key "$NANSEN_API_KEY"`. The parser leaves the expanded key in `extraArgs[0]`, and this template writes it verbatim into the error envelope. With `--json`, the full key lands on stdout, which CI and agent runners commonly persist. I reproduced this at the PR head. Please use a fixed error saying that `mcp verify` takes no positional arguments without interpolating the argument. Heuristic masking is less safe than never echoing it.

Codier (changes requested):
Requesting changes for the credential exposure above. Separately, the current head is unattributed on GitHub and this agent-authored PR has no `## AI execution metadata` section. Please rewrite the commits under `hulk-linus[bot]` and add the required provider, model, thinking level, token usage, and attribution scope before merge.

Codier (dismissed):
The code issue is fixed and the exact-head security review is clean. The remaining blockers are unchanged: GitHub still resolves the current head as unattributed, and the PR body still has no `## AI execution metadata` section. Please rewrite the agent-authored commits under `hulk-linus[bot]` and add the required execution metadata before merge.

Codier (approved):
The credential-output issue is fixed. The exact-head security review and regression tests pass.

PR #598, opened 2026-09-07
fix: verify limit-order deposit/cancel outcome before signing

kome12:
## Summary
- `assertSolanaInstructionsSafe` never classified SPL `Transfer`/`TransferChecked`, so an API-crafted limit-order deposit or cancel transaction could move unrelated funds out of the wallet (e.g. a plain `SystemProgram.transfer` to an attacker address) and still pass the only pre-signing guard on that path.
- Adds balance-delta outcome verification for both flows, mirroring the existing swap-execution path: a deposit must move exactly the requested input amount out of the wallet (± fee/rent slack for native SOL) and no other token may leave; a cancel must not move any token out of the wallet at all.
- Verification degrades gracefully (warns and proceeds) when no simulation RPC is available, but fails closed — refuses to sign — on a real outcome mismatch or an in-simulation revert.
- This closes the *magnitude* gap (unbounded/partial drains). Binding the deposit's *destination* to the vault's own token account is tracked separately as a fast-follow, since balance-delta simulation of the wallet's own accounts can't see where funds landed once the amount matches.

## Test plan
- [x] `npm test` — all unit tests pass, including 14 new tests for the two pure outcome asserters (upper/lower bound, sibling-token drain, native-SOL fee/rent slack, fail-closed on non-integer deltas) and 3 new integration tests proving the create/cancel command handlers actually refuse to sign a simulated drain (and still degrade gracefully on a simulation RPC error)
- [x] `npm run lint`
- [x] `/security-review` on the diff — no findings
- [x] Ran the real `node ./src/index.js trade limit-order create` binary end-to-end against a local mock trading-api + Solana RPC: confirmed a clean deposit is verified and the order is created, and a simulated drain (the exact bug this PR fixes) is refused before signing
- [x] Manual live e2e round-trip on a funded Solana wallet — SOL→USDC and USDC→SOL, both `create` + `cancel`. Every deposit and cancel printed `✓ outcome verified (via simulateTransaction)`: the new check ran (did not degrade to warn-and-proceed) and did not false-block a legitimate deposit/withdrawal. Order value had to clear Jupiter's $10 minimum (0.1 SOL / 11 USDC); funds fully returned on cancel (input balances restored).


kome12:
@nansen-pr-reviewer re-review

Codier (approved):
Hey Ko, I read through the whole diff and ran the tests locally (337 pass, lint clean). Approving.

The fix is solid. The two new asserters mirror `assertSolanaSwapOutcome` closely enough that I'm confident the bounds are right. The degrade policy matches the swap path exactly: warn and proceed on RPC errors, refuse to sign on a real mismatch or a sim revert. The cancel check binds to the full remaining refund instead of just checking for a positive inflow, which closes the dust-refund redirect. That's a real gap and this fixes it.

Two things I'd want to clean up before or as a fast-follow:

1. `remainingRefundAmount` returns `undefined` for both "can't parse the amounts" and "order is fully filled." A user who tries to cancel a filled order that's still showing as active gets "could not determine its remaining refund amount" which doesn't tell them what happened. Worth distinguishing those two cases so the error is actionable. Not a security issue, fail-closed is correct either way.

2. There's a narrow race between the order lookup (step 2) and the cancel request (step 3). If a partial fill lands in between, the actual refund will be less than what you computed, and the check fails closed on a legitimate cancel. The user can retry and it'll work. I think that's acceptable, but worth a comment so nobody "fixes" it by loosening the check.

The destination gap (right amount, wrong address) is called out in the PR body and the JSDoc. That's the right scope for this PR.

One question: you said the manual live e2e round-trip isn't done yet. Are you planning to run that before merge, or merge and verify after? Run it first if you can. This touches the signing path on real funds.

kome12:
> Hey Ko, I read through the whole diff and ran the tests locally (337 pass, lint clean). Approving.
> 
> The fix is solid. The two new asserters mirror `assertSolanaSwapOutcome` closely enough that I'm confident the bounds are right. The degrade policy matches the swap path exactly: warn and proceed on RPC errors, refuse to sign on a real mismatch or a sim revert. The cancel check binds to the full remaining refund instead of just checking for a positive inflow, which closes the dust-refund redirect. That's a real gap and this fixes it.
> 
> Two things I'd want to clean up before or as a fast-follow:
> 
> 1. `remainingRefundAmount` returns `undefined` for both "can't parse the amounts" and "order is fully filled." A user who tries to cancel a filled order that's still showing as active gets "could not determine its remaining refund amount" which doesn't tell them what happened. Worth distinguishing those two cases so the error is actionable. Not a security issue, fail-closed is correct either way.
> 2. There's a narrow race between the order lookup (step 2) and the cancel request (step 3). If a partial fill lands in between, the actual refund will be less than what you computed, and the check fails closed on a legitimate cancel. The user can retry and it'll work. I think that's acceptable, but worth a comment so nobody "fixes" it by loosening the check.
> 
> The destination gap (right amount, wrong address) is called out in the PR body and the JSDoc. That's the right scope for this PR.
> 
> One question: you said the manual live e2e round-trip isn't done yet. Are you planning to run that before merge, or merge and verify after? Run it first if you can. This touches the signing path on real funds.

I'd do a fast-follow for clean up.

I did a manual live e2e round trip just now. Updated the PR description as well.

PR #599, opened 2026-09-07
fix(x402): bind scheme, network, and signing domain before signing

kome12:
## Summary

- Refuse any payment scheme other than `"exact"` at the policy layer, blocking all signing paths before any key material is touched
- Key the SVM token table by exact CAIP-2 id (`solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp`) so unknown `solana:*` network strings resolve to no token and are refused
- Derive the EIP-712 `domain.chainId` only from the validated CAIP-2 `network` field; reject a conflicting remote `extra.chainId` instead of letting it silently override
- Harden `getSolanaRpcUrl` to throw on unknown networks rather than falling back to mainnet

## Test plan

- [ ] `npm test` passes (2730 tests)
- [ ] `npm run lint` passes
- [ ] New unit tests cover: unsupported/missing scheme refused, unknown `solana:*` CAIP-2 id refused, conflicting `extra.chainId` throws, missing `extra.name`/`extra.version` throws, `getSolanaRpcUrl` throws on unknown network
- [ ] Existing happy-path tests for Base USDC, Solana mainnet USDC, and BSC tokens all still pass

### DIFF
--- a/.changeset/x402-scheme-network-binding.md
+++ b/.changeset/x402-scheme-network-binding.md
@@ -1,0 +1,5 @@
+---
+"nansen-cli": patch
+---
+
+x402 auto-pay now signs only for supported schemes and recognized payment networks (exact Solana mainnet CAIP-2 binding), and derives the EIP-712 chain id from the validated network.
--- a/.changeset/x402-scheme-network-binding.md
+++ b/.changeset/x402-scheme-network-binding.md
@@ -2,4 +2,4 @@
 "nansen-cli": patch
 ---
 
-x402 auto-pay now signs only for supported schemes and recognized payment networks (exact Solana mainnet CAIP-2 binding), and derives the EIP-712 chain id from the validated network.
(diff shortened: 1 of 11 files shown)


Codier (approved):
Approving. One small non-blocking suggestion. The "Paid Access Rails" paragraph in AGENTS.md still lists the policy guard's checks as (network, asset) plus the amount cap plus the payTo allowlist, and doesn't mention the new scheme gate or the exact Solana CAIP-2 binding. Since AGENTS.md steers agent work on this repo, worth a one-line update.

kome12:
> Approving. One small non-blocking suggestion. The "Paid Access Rails" paragraph in AGENTS.md still lists the policy guard's checks as (network, asset) plus the amount cap plus the payTo allowlist, and doesn't mention the new scheme gate or the exact Solana CAIP-2 binding. Since AGENTS.md steers agent work on this repo, worth a one-line update.

I'll do a fast-follow to add the one line update!

PR #600, opened 2026-09-08
fix(cache): isolate response cache by credential and request context

kome12:
## Summary

The opt-in response cache (`--cache`) derived its key from only the endpoint and request body. Because the key ignored who made the request and where it was sent, clients sharing the same `$HOME` (and therefore the same cache directory) could read each other's cached responses — e.g. a response cached under one API key or base URL could be served to a request made under a different one, with no HTTP request.

This change folds every request property that affects authorization or response semantics into the cache key:

- **Base URL / origin** and **HTTP method**
- A **hashed identity** — the API key plus auth-relevant headers (`apikey`, `authorization`, `payment-signature`), digested with SHA-256 so no raw credential appears in a cache filename or the key's pre-image
- Endpoint and body (unchanged)

The key digest also moves from MD5 to SHA-256.

## Notes

- Caching is opt-in and off by default; this only affects `--cache` runs.
- No migration needed — changing the key scheme means old-format entries are never looked up again (they simply become invisible to the new keying), rather than being actively expired by the TTL check. They're inert, harmless leftovers that `nansen cache clear` removes; nothing reads or serves them.
- Cache file contents are unchanged; identity lives only in the key derivation.

## Test plan

- [x] `npm test` passes (2728 passed, 9 skipped)
- [x] `npm run lint` passes
- [x] New regression tests: same endpoint + body under a different identity, base URL, or HTTP method is a cache miss

### DIFF
--- a/.changeset/cache-isolation.md
+++ b/.changeset/cache-isolation.md
@@ -1,0 +1,8 @@
+---
+"nansen-cli": patch
+---
+
+Isolate the response cache by credential and request context so cached
+responses can no longer be shared across different API keys or API origins that
+use the same cache directory. Cache keys now include the base URL, HTTP method,
+and a hashed form of the effective credentials, and use SHA-256.
--- a/src/__tests__/api.test.js
+++ b/src/__tests__/api.test.js
@@ -9,7 +9,7 @@ import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
 import { describe, it, expect, beforeAll, vi } from 'vitest';
-import { NansenAPI, ErrorCode } from '../api.js';
+import { NansenAPI, ErrorCode, clearCache } from '../api.js';
 
 const LIVE_TEST = process.env.NANSEN_LIVE_TEST === '1';
 const API_KEY = process.env.NANSEN_API_KEY || 'test-key';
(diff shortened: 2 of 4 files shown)


gulshngill (dismissed):
## End-to-end review — approve

I reviewed the full user-facing flow: CLI command wiring → request/business logic → response/cache handling → error paths → tests, and validated the change with a local mock API and the merged main+PR tree.

### Verified behavior

- The original credential-isolation bug reproduces on `main`: after key A populates the response cache, key B can receive key A's cached response.
- PR #600 fixes this: key A and key B produce separate cache entries and responses.
- Base-URL/request-context isolation was also exercised.
- The merged tree is clean with no merge conflicts.

### Findings

1. **Medium — test coverage gap (non-blocking).** The fix guards the two cache call sites in `src/api.js` (`703–704` and `933–934`) so a nullable `cacheContext` cannot reach cache-key generation, but there is no regression test directly exercising those API call sites with a missing identity/context. The existing cache-isolation tests validate the core behavior, so this is a coverage recommendation rather than a correctness blocker.

2. **Low — PR hygiene/description inconsistency (non-blocking).** There is a small inconsistency in the change description/documentation that should be cleaned up separately; it does not affect runtime behavior.

### Validation

- Full test suite: **2,733 passed, 9 skipped**
- Targeted cache tests: **19 passed**
- Lint: **passed**
- CLI run against a local mock API: **passed**
- Main-vs-PR reproduction confirmed the cache leak is fixed

No blocking correctness, security, or merge issues found. Approving.

kome12:
Addressed both:

1. Test coverage gap — added an integration-level regression test that exercises the `useCache && cacheContext` guards through `NansenAPI.request()` itself (two different API keys, same endpoint/body: cross-key miss, then same-key hit), not just the underlying cache helpers. (141766b)
2. Description inconsistency — the "age out on the existing 5-minute TTL" line was wrong: old-format entries are never looked up again under the new key scheme, so the TTL check (which only fires on a lookup) never touches them. They're just inert orphaned files until `nansen cache clear`. Corrected in the PR description.

PR #608, opened 2026-09-09
Fix Solana trade e2e gas reserve

kome12:
## Summary
- raise the Solana reserve left after the cross-chain e2e reverse leg to 0.015 SOL
- document that the reserve must remain above the CLI Solana gas floor for the following SOL/USDC suite

## Tests
- npm test
- npm run lint

Not run: npm run test:trade, because it requires funded wallets and real network broadcasts.


kome12:
Addressed: updated the reserve comment to cross-reference validateGasBalance() and MIN_GAS_AMOUNTS.solana directly. Re-ran npm run lint.

PR #615, opened 2026-09-10
fix: WalletConnect EVM signing accepted a session approved for a different chain

### LINKED ISSUE #614: WalletConnect EVM signing accepts a session approved for a different chain
**What happens**

`nansen transfer --wallet walletconnect` and `nansen trade execute` (EVM, `--wallet walletconnect`) accept any connected WalletConnect session as valid for the target chain, without checking that the session was actually approved for that specific chain.

**Root cause**

`getWalletConnectAddress('evm')` in `src/walletconnect-trading.js` matches any account whose CAIP-2 chain tag starts with `eip155:`, regardless of which specific `eip155:<chainId>` it was approved for:

```js
if (chainType === 'evm') {
  const evmAccount = accounts.find(a => a.chain?.startsWith('eip155:'));
  return evmAccount?.address || null;
}
```

Because EVM addresses are identical across every EVM chain, a WalletConnect session connected only to, say, Ethereum mainnet (`eip155:1`) is silently treated as valid for signing/broadcasting on Base (`eip155:8453`) or any other EVM chain. Nothing downstream can catch this: `assertQuoteMatchesRequest` and the other quote/request-intent binding checks only compare the resolved address against the persisted request, not the chain the session was actually approved for.

The equivalent Solana branch in the same function already does an exact match (`chain === SOLANA_MAINNET_CHAIN`) instead of a prefix match, so this is an inconsistency within the function itself, not just a general design gap.

**Impact**

A user who has WalletConnect connected to one EVM chain, then runs `nansen transfer` or `nansen trade execute` targeting a different EVM chain, gets no error -- the wrong-chain session is used to build and sign the request. Depending on the connected wallet's own chain handling this can range from a confusing failure to an actual cross-chain signature/broadcast the user did not intend.

**Where**

- `src/walletconnect-trading.js`: `getWalletConnectAddress(chainType)`
- `src/transfer.js`: `sendTokensViaWalletConnect` calls `getWalletConnectAddress()` with no chain argument at all
- `src/trading.js`: three call sites in the `trade execute`/`quote` WalletConnect path call `getWalletConnectAddress(chainType)` without a chain ID

**Suggested fix**

Give `getWalletConnectAddress` an optional `chainId` parameter that, when provided, requires an exact `eip155:<chainId>` match (mirroring the existing Solana exact-match idiom in the same function), and pass the target chain's ID at every EVM call site that resolves a WalletConnect address before signing or broadcasting.

I have a fix ready and will open a PR referencing this issue.

teyrebaz33:
Fixes #614.

## Problem

`getWalletConnectAddress('evm')` matched any connected account whose CAIP-2 chain tag started with `eip155:`, regardless of which specific chain it was approved for. Because EVM addresses are identical across chains, a WalletConnect session connected only to Ethereum mainnet would be silently accepted as valid for a Base (or any other EVM) transaction. Nothing downstream -- including the quote/request-intent binding checks -- could catch this, since they only compare addresses, not chains. The equivalent Solana branch in the same function already did an exact match against Solana mainnet, so this was an inconsistency within the function itself.

## Fix

`getWalletConnectAddress` now accepts an optional `chainId` and, when given, only returns an account the session has actually approved for that exact chain (`eip155:<chainId>`), mirroring the existing Solana exact-match idiom.

Every place that resolves a WalletConnect EVM address before signing or broadcasting a real transaction now passes the target chain ID and fails closed with a clear error instead of proceeding with a wrong-chain session:

- `nansen transfer`'s `sendTokensViaWalletConnect`
- `nansen trade execute`'s quote-building step
- `nansen trade execute`'s pre-execute wallet-match check
- `nansen trade execute`'s immediate pre-signing check (the last line of defense right before a transaction is signed/broadcast)

The last two are deliberately both kept (not collapsed into one check) -- a new regression test confirms this closes the gap where a session could disconnect or switch chains between the pre-execute check and the moment of signing.

## Testing

- `npx eslint .` -- clean
- Full test suite: 2860 passed, 16 skipped (no regressions)
- New regression tests added covering: exact chainId match, a session on a different chain being rejected instead of matched, backward compat when chainId is omitted, `nansen transfer` scoping its lookup to the target chain, and `nansen trade execute` rejecting both a session on the wrong chain up front and one that switches chains mid-execute

https://claude.ai/code/session_0141eH4NcQy4GXgusCGbG1pm

### DIFF
--- a/.changeset/walletconnect-chain-scoped-address.md
+++ b/.changeset/walletconnect-chain-scoped-address.md
@@ -1,0 +1,9 @@
+---
+"nansen-cli": patch
+---
+
+Fix `nansen transfer`/`nansen trade execute` with `--wallet walletconnect` using any connected EVM account instead of verifying the WalletConnect session is actually approved for the chain being signed/broadcast on.
+
+`getWalletConnectAddress('evm')` matched any account whose CAIP-2 chain tag started with `eip155:`, regardless of which specific chain it was approved for. Because EVM addresses are identical across chains, a session connected only to Ethereum mainnet would be silently used to sign a transaction destined for Base (or vice versa) -- nothing downstream (including the quote/request-intent binding checks) could catch this, since they only compare addresses, not chains.
+
+`getWalletConnectAddress` now accepts an optional `chainId`, and when given, only returns an account the session has approved for that exact chain (mirroring the mainnet-only exact match already used for Solana in the same function). Every place that resolves a WalletConnect EVM address before signing or broadcasting a real transaction now passes the target chain ID and fails closed with a clear error instead of proceeding with a wrong-chain session: `nansen transfer`'s `sendTokensViaWalletConnect`, and `nansen trade execute`'s quote-building, pre-execute wallet-match check, and immediate pre-signing check for its EVM WalletConnect swap path.
--- a/src/__tests__/trading.test.js
+++ b/src/__tests__/trading.test.js
@@ -1577,6 +1577,141 @@ describe('WalletConnect execute support', () => {
     vi.restoreAllMocks();
   });
 
+  it('scopes the WalletConnect address lookup to the quote chain before signing', async () => {
+    const addrSpy = vi.spyOn(wcTrading, 'getWalletConnectAddress').mockResolvedValue('0x742d35Cc6bF4F3f4e0e3a8DD7e37ff4e4Be4E4B4');
+    vi.spyOn(wcTrading, 'sendTransactionViaWalletConnect').mockResolvedValue({ txHash: '0xmocktx' });
+
+    const originalFetch = global.fetch;
+    global.fetch = vi.fn(async () => ({
+      status: 200,
+      text: () => Promise.resolve(JSON.stringify({ result: { status: '0x1', blockNumber: '0x100' } })),
+      json: () => Promise.resolve({ result: { status: '0x1', blockNumber: '0x100' } }),
+    }));
+
+    const quoteId = saveQuote({
+      success: true,
+      quotes: [{
+        aggregator: 'lifi',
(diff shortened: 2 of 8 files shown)


teyrebaz33:
Thanks -- confirmed and fixed in dc18351, plus the sibling copy of the same defect I found while checking this.

One correction on reachability: `sendTokensViaWalletConnect` is only reached through `wallet send`, which hard-validates `--chain` against `['evm', 'solana', 'ethereum', 'base']` in `src/wallet.js` before calling `sendTokens` -- all four resolve correctly in `CHAIN_IDS`, so the `ronin`/`sei`/etc. scenario can't actually reach this code today. Fixed it anyway (same defect class, cheap fix, and a real hardening against a future loosened validation or a new caller), using the exact idiom already in this file at the Privy EVM path (line 650-651) rather than the wording suggested above, for consistency.

Also found and fixed the same `CHAIN_IDS[chain] || 1` pattern in `buildEvmTransaction` (the local-wallet EVM signing path), which had the identical gap.

teyrebaz33:
Addressed all three in bc17617 -- comment-only, no functional change:

1. Added a note above `EVM_CHAIN_IDS` in `src/chain-ids.js` pointing out the trap you flagged.

2. Added a comment at both `trading.js` call sites explaining why `chainId` is EVM-only and why the ternary can't be collapsed to an unconditional pass without breaking Solana.

3. Reworded the JSDoc's closing note in `walletconnect-trading.js` -- confirmed via a full `grep` that every remaining no-`chainId` call is either Solana or truly chain-agnostic, so the old "callers that don't yet sign/broadcast" phrasing was outdated.

kome12 (approved):

PR #627, opened 2026-09-16
Harden EVM bridge deposit validation

kome12:
## Summary

- reject non-USDC Base-to-Hyperliquid bridge deposit quotes before quote creation
- add execute-path coverage for tampered EVM bridge plans
- return coded, actionable bridge validation errors for malformed amount anchors and over-cap approvals
- normalize accepted deposit calldata before signing

## Testing

- npm test
- npm run lint


kome12:
@nansen-pr-reviewer re-review

Codier (changes requested):
Please fix the bridge-step ordering guard.

[`src/bridge.js:1018`](https://github.com/nansen-ai/nansen-cli/blob/7d6d4682d3b6fa9751fc3d6642ad16e2928317bf/src/bridge.js#L1018) checks the number of approve and deposit items but not their order. A tampered quote can put the deposit before the approval. If that deposit succeeds using an existing router allowance, the trailing approval restores a capped allowance that no remaining deposit consumes. This defeats the guard against a standalone router allowance.

Require every incomplete approval to precede the deposit, including items bundled within a step. Also reject zero-amount deposits. Please add cases for `[deposit, approve]`, `[deposit, complete-approve, approve]`, and zero-amount deposits.

Codier (approved):
Ordering finding is fixed. [`preflightEvmBridgeSteps`](https://github.com/nansen-ai/nansen-cli/blob/3d972ee2e072a5ea08a418788ec3385772558863/src/bridge.js#L1015) now tracks `sawDeposit` and rejects any incomplete approve that follows the deposit. Because the loop is over `step.items` rather than steps, an approve bundled after the deposit inside a single step is caught too, and the `status === 'complete'` skip leaves `[deposit, complete-approve, approve]` still tripping the guard. Zero-amount deposits are rejected at [`src/bridge.js:976`](https://github.com/nansen-ai/nansen-cli/blob/3d972ee2e072a5ea08a418788ec3385772558863/src/bridge.js#L976), before the anchor lookup.

Tests cover the three requested shapes plus the zero-amount deposit. `broadcasts the normalized deposit calldata, not the dirty server bytes` is the one that matters most: `preflightEvmBridgeSteps` discards the validator's return value, so it pins that the signing path re-validates at [`src/bridge.js:1210`](https://github.com/nansen-ai/nansen-cli/blob/3d972ee2e072a5ea08a418788ec3385772558863/src/bridge.js#L1210) and signs `bound.data` instead of the server's bytes.

On the remaining bot finding, I'd leave the zero-deposit message as it is. "Refusing to sign. Request a new quote." is the pattern 23 other refusals in this file already use. The approve-branch catch strips that phrase only because `encodeApproveCalldata`'s own message ends with it and would otherwise stack two directives; the hand-written deposit message isn't doubling anything.

Checks: CI green on this head (lint, test 20/22/24, Aikido). Locally `npm run lint` clean and `npm test` 2929 passed, with one unrelated failure in `src/__tests__/ens.test.js` — a real ENS lookup that times out without network egress in my sandbox, in a file this PR does not touch and which passes in CI.

PR #629, opened 2026-09-16
fix(profiler): validate trace depth before traversal

kriss39:
## Summary

`profiler trace --depth` currently uses `parseInt()`, so malformed values like `abc` or `Infinity` can turn into `NaN` and reach the traversal logic.

This PR adds proper validation before the trace starts.

### Changes

- reject malformed, fractional, repeated and valueless `--depth`
- keep the existing 1–5 clamping behavior
- fail before making counterparty API calls
- update the schema to mark `depth` as an integer
- add regression tests

This avoids invalid input silently affecting the BFS depth limit.


src/cli.js
```diff
@@ -606,6 +606,31 @@ export async function batchProfile(api, params = {}) {
...
+    ? Number(raw)
+    : NaN;
+  if (!Number.isSafeInteger(value)) {
+    throw new NansenError(
+      `--depth must be a safe integer; received: ${String(raw)}`,
+      ErrorCode.INVALID_PARAMS,
+    );
+  }
+  return Math.max(1, Math.min(value, 5));
+}
```
kome12, line 653: `normalizeTraceDepth` duplicates most of [parseNonNegativeSafeIntegerOption](https://github.com/kriss39/nansen-cli/blob/b802bca7ad5352ba03226fdbc64e97a3f4496d03/src/cli.js#L223) (array check, empty-string check, NaN/safe-integer check, bare-flag check). Could we avoid copy-pasting that? Straight reuse of `parseNonNegativeSafeIntegerOption` won't quite work as-is since it rejects negative values instead of clamping them to 1, which changes the documented clamping behavior this PR is trying to preserve. Could you pull the shared parse-and-validate logic into a lower-level helper (no bound opinion), have `parseNonNegativeSafeIntegerOption` call it with a < 0 check, and have depth call it too with its own `Math.max(1, Math.min(value, 5))` clamp? That keeps the validation logic in one place without forcing depth to adopt reject-on-negative semantics.

kome12 (commented):
Thanks for finding and fixing the issue! I just have 1 comment to reduce some of the duplicate code.

src/cli.js
kome12: So maybe something like this?
```
// low-level: parse + validate, no bound opinion
function parseSafeIntegerOption(name, options, flags, defaultValue) {
  if (flags[name]) {
    throw new NansenError(`--${name} requires a safe integer value`, ErrorCode.INVALID_PARAMS);
  }
  if (options[name] === undefined) return defaultValue;

  const rawValue = options[name];
  if (Array.isArray(rawValue)) {
    throw new NansenError(`--${name} may only be specified once`, ErrorCode.INVALID_PARAMS);
  }
  if (typeof rawValue === 'string' && rawValue.trim() === '') {
    throw new NansenError(`--${name} requires a safe integer value`, ErrorCode.INVALID_PARAMS);
  }
  const value = typeof rawValue === 'string' || typeof rawValue === 'number' ? Number(rawValue) : NaN;
  if (!Number.isSafeInteger(value)) {
    throw new NansenError(`--${name} must be a safe integer; received: ${String(rawValue)}`, ErrorCode.INVALID_PARAMS);
  }
  return value;
}

function parseNonNegativeSafeIntegerOption(name, options, flags, defaultValue) {
  const value = parseSafeIntegerOption(name, options, flags, defaultValue);
  if (value < 0) {
    throw new NansenError(`--${name} must be a non-negative safe integer; received: ${value}`, ErrorCode.INVALID_PARAMS);
  }
  return value;
}
```

kome12 (approved):

PR #630, opened 2026-09-16
fix(cli): validate analytics days options

kriss39:
## Summary

Several analytics commands currently use `parseInt()` for `--days`.

That means values like `7abc` become `7`, `2.5` becomes `2`, and some invalid values can reach the date-range logic.

This PR makes `--days` validation stricter and consistent across the CLI.

### Changes

- reject malformed, negative, fractional, repeated and valueless values
- reject values that cannot produce a valid date range
- keep valid values such as `--days 0`
- apply the validation only to commands that support `--days`
- update the schema and add regression tests

This makes invalid input fail early instead of being silently changed.


kome12 (commented):
Thanks for fixing this as well! The logic is fine, just small nits that the [PR reviewer](https://github.com/nansen-ai/nansen-cli/pull/630#issuecomment-5696126540) addressed will just clean it up more.

kome12:
@kriss39 Can you rebase your branch from the latest main? I think it's just some staleness that's causing the issue.

kriss39:
Rebased the branch onto the latest `main` and cleaned up the stale merge state. I also fixed the test path naming issue caught by lint. CI is running again now.

kome12 (approved):

PR #631, opened 2026-09-16
fix(pm): validate numeric screener filters

Bruce039:
## Summary

Prediction-market screener filters currently use `Number()` directly, so malformed values like `NaN` or `Infinity` can reach the request payload.

This adds validation before the API call.

### Changes

- reject malformed, non-finite, repeated and valueless numeric filters
- keep valid finite values unchanged
- cover both market and event screeners
- add regression tests

This prevents invalid numeric input from being silently serialized or forwarded.


kome12 (commented):
One of your tests is failing. Can you look into fixing that?

PR #632, opened 2026-09-16
fix(cli): reject invalid date options

Bruce039:
## Summary

Invalid `--date` values currently fall back to the rolling `--days` range instead of failing, which can run a different query than the user intended.

This validates the date before the API call.

### Changes

- reject malformed and valueless `--date`
- validate real calendar dates
- keep the existing `--days` fallback when `--date` is omitted
- support valid date-range objects
- add regression tests

Bad date input now fails clearly instead of silently falling back.


kome12 (commented):
The feedback from the pr-reviewer [here](https://github.com/nansen-ai/nansen-cli/pull/632#issuecomment-5705356609) makes sense. It would be better to address this as well.

Bruce039 (commented):
Thanks, that makes sense. I checked the failing case again and agree that a `{ from }` date-range object should be normalized the same way as the single-date path.

I’ll update the implementation to return `{ from, to: to ?? from }` so `to` defaults to `from` when it is omitted, and keep the regression test for this case.

kome12 (approved):

PR #636, opened 2026-09-17
fix(cli): validate string options before .split()/.trim() calls

kome12:
## Summary

- `parseArgs` runs `JSON.parse` on option values, so `--flag true` / `--flag false` / `--flag null` become JS primitives rather than strings. Several CLI handlers called `.split(',')`, `.trim()`, `.indexOf()`, `.lastIndexOf()`, or `.toLowerCase()` directly on an option without checking `typeof`, so a JSON-primitive-looking value crashed with a raw, unhandled `TypeError` instead of an actionable `INVALID_PARAMS` error.
- Adds a shared `splitCsvOption()` guard in `src/cli.js` and applies it at each unguarded call site: `profiler batch --include`, `perp screener --sectors-filter`/`--sm-label-filter`/`--trader-label-filter`, and `prediction-market market-screener`/`event-screener --tags`.
- Hardens `parseFields()` (`--fields`) the same way, and narrows its early-return so falsy non-string primitives (`--fields false` / `--fields null`) hit the type guard instead of silently returning an unfiltered result.
- Hardens `src/commands/alerts.js`: `parseChains`/`parseTokens`/`parseSubjects` (used across all alert types) and the `alerts list --token-address`/`--chain` filters now reject non-string option values instead of crashing.
- Adds regression tests for every fixed flag, confirming a clean `INVALID_PARAMS` error instead of a crash, plus coverage for `web search --query` passed as a JSON object or an array of non-strings (behavior was already correct, just untested).

## Test plan

- [x] `npm test` — all suites pass
- [x] `npm run lint` — clean
- [x] Manually reproduced each fixed flag locally and confirmed a clean `INVALID_PARAMS` message instead of a raw `TypeError`
- [x] Manually confirmed a valid comma-separated value for each flag still works end-to-end


kome12:
@nansen-pr-reviewer re-review

PR #637, opened 2026-09-17
fix: align counterparties-batch chain allowlist with the endpoint

gulshngill:
Follow-up to the non-blocking note on the final review of **#624** ([review](https://github.com/nansen-ai/nansen-cli/pull/624#pullrequestreview-5231017813)). Does not touch #624.

- nansen-cli **#624** — https://github.com/nansen-ai/nansen-cli/pull/624 (merged, `9cfe531`)
- nansen-ra **#3550** — https://github.com/nansen-ai/nansen-ra/pull/3550 (open, defines `ADDRESS_COUNTERPARTIES_BATCH_CHAINS`)

## The mismatch

`addressCounterpartiesBatch` validated `--chain` against `['all', 'solana', ...EVM_CHAINS]`. `EVM_CHAINS` is this CLI's *address-format and ENS* set (`src/chain-ids.js`), not a profiler chain set, so it disagreed with the endpoint in both directions — and separately with the MCP side.

| `--chain` | CLI at `9cfe531` | nansen-ra #3550 | Endpoint (`ProfilerChain`) |
|---|---|---|---|
| `scroll`, `ronin` | accepted → request sent | rejected | **not in the enum → 422** |
| `bitcoin`, `tron`, `sui`, `ton`, `near`, `injective`, `mantra`, `robinhood` | rejected | accepted | **accepted** |
| `arc`, `starknet` | rejected | rejected | **accepted** |
| `bsc` | rejected | normalised to `bnb` | `bnb` only |

So the CLI both sent two chains that can only 422 and refused ten chains the endpoint serves — the latter a regression against pre-#624 behaviour, where those reached the server through `validateAddress`'s permissive fallback.

## Authoritative contract and rationale

The endpoint's request model is `ProfilerAddressCounterpartiesBatchRequest`, whose `chain` is `$ref: ProfilerChain` and whose body is `additionalProperties: false`. A value outside that enum is rejected by the server, so **the enum is the contract**. Read from the live spec at `https://api.nansen.ai/openapi.json` (the same URL `src/cost-cache.js` already uses):

```
all, arbitrum, arc, avalanche, base, bitcoin, bnb, ethereum, hyperevm,
injective, iotaevm, linea, mantle, mantra, monad, near, optimism, plasma,
polygon, robinhood, sei, solana, sonic, starknet, sui, ton, tron
```

`COUNTERPARTIES_BATCH_CHAINS` in `src/api.js` is now exactly that list, as a **batch-specific constant with a comment saying why it is not `EVM_CHAINS`** — the shape the reviewer asked for, so the intent is stated rather than implied by a reused list.

Three deliberate calls:

1. **`bsc` → `bnb`, not rejected.** The enum only knows `bnb`, but `bsc` is the common spelling and the one the API echoes back in responses (the `--chain bnb` quirk already documented in `CLAUDE.md`). Normalising matches #3550 and is strictly more permissive than before; the normalised `bnb` is what reaches the wire *and* what address validation runs against, so a `bsc` batch is still format-checked as EVM. Scoped to this command — no other CLI command accepts `bsc` today, and widening that is out of scope here.
2. **Non-EVM chains are intentionally supported, not intentionally rejected.** The endpoint serves them, and #624 never decided otherwise — it rejected them as a side effect of sourcing the list from `EVM_CHAINS`.
3. **Admitting them does not re-weaken validation.** `validateAddress` only has patterns for EVM, Solana and Bitcoin; for `arc`, `injective`, `mantra`, `near`, `robinhood`, `starknet`, `sui`, `ton` and `tron` it falls through to "any non-empty string", which is exactly the `--file` hole [P3 of the `2a1ea27` review](https://github.com/nansen-ai/nansen-cli/pull/624#pullrequestreview-5223455434) closed. So a named chain now runs `requireValidAddress` **and** an `ADDRESS_SHAPED` floor (3–128 chars, no whitespace, alphanumerics plus `.`/`-`/`_` — enough for bech32, NEAR and TON). Every valid EVM, Solana and Bitcoin address clears it, so it never fires on a chain that was already strictly validated; a stray `--file` line still cannot reach the wire. `validateAddress`'s global permissive fallback is untouched: tightening it would change every command, which is a separate change.

**Divergence left open on purpose:** #3550 omits `arc` and `starknet`, describing its list as "the intersection of the upstream `ProfilerChain` enum and the address formats this client validates". The endpoint accepts both, so the CLI accepts both. If the MCP side prefers the intersection framing, `arc`/`starknet` are the two rows to reconcile there — the CLI is the superset and matches the server.

## Files changed

| File | Change |
|---|---|
| `src/api.js` | `COUNTERPARTIES_BATCH_CHAINS` (exported) + `ADDRESS_SHAPED` + `resolveCounterpartiesBatchChain()`; method validates and sends `resolvedChain`, and a named chain gets the shape floor after `requireValidAddress`. |
| `src/__tests__/api.test.js` | See below. |
| `src/schema.json` | `counterparties-batch` → `chain` description enumerates the supported set and the `bsc` alias. |
| `skills/nansen-wallet-profiler/SKILL.md` | Same list next to the `counterparties-batch` example. |
| `.changeset/counterparties-batch-chain-allowlist.md` | `patch`. |

`src/cli.js` is unchanged — `chain` is passed straight through and normalised in the API layer.

## Tests

Added to `describe('addressCounterpartiesBatch')`, next to the existing guards:

- the allowlist pinned against the spec enum (both directions of the mismatch in one assertion);
- `scroll` and `ronin` rejected, 0 fetches (`it.each`);
- `tron` sent with `chain: "tron"`;
- a TON address not lowercased (case-sensitive non-EVM chain);
- an arbitrary `--file`-style line rejected on `tron` with 0 fetches — the P3 regression guard;
- `bsc` sent as `bnb`, with a checksum-cased address still lowercased;
- a malformed address on `bsc` reported against `bnb` as an EVM address.

The pre-existing `should reject an unsupported named chain` case used `bsc`, which is now valid; it uses `zksync` instead and keeps asserting 0 fetches.

## Validation

```
npx vitest run src/__tests__/api.test.js -t addressCounterpartiesBatch
  → Tests  23 passed | 214 skipped (237)

npm test
  → Test Files  72 passed (72)
    Tests  2951 passed | 9 skipped (2960)

npm run lint        → exit 0, no findings
git diff --check    → clean
node -e "JSON.parse(fs.readFileSync('src/schema.json'))" → ok
```

Stubbed-fetch matrix driving `addressCounterpartiesBatch` at this head, no network — the inversion the reviewer reported, resolved:

```
scroll, ronin                  -> rejected (0 fetch)
zksync                         -> rejected (0 fetch)
bsc, bnb                       -> sent (1 fetch, chain "bnb")
bitcoin, tron, sui, ton, near  -> sent (1 fetch)
injective, mantra, robinhood   -> sent (1 fetch)
arc, starknet                  -> sent (1 fetch)
all, solana                    -> sent (1 fetch)
tron + "not an address, just a line" -> rejected (0 fetch)
```

No live API calls were made from this branch; the spec was fetched read-only to derive the enum.


gulshngill:
## Addressed: `ADDRESS_SHAPED` rejects raw TON addresses (medium)

Fixed in `427f42f`. Took a third option rather than A or B — a chain-aware raw-TON check, so the floor for every other chain stays exactly as it was.

### Why not option A

Adding `:` to `ADDRESS_SHAPED` fixes TON by loosening the floor for all 27 chains, including the ones where no address ever contains a colon, and it still would not have accepted `-1:…` — the leading `-` fails the anchored first character. Option B keeps the floor tight but leaves valid addresses rejected, which is the actual bug.

### What changed — `src/api.js`

```js
/**
 * TON's raw address form: a workchain id, ":", then 64 hex characters —
 * `0:` for the basechain, `-1:` for the masterchain. ...
 */
const TON_RAW_ADDRESS = /^-?\d{1,10}:[0-9a-fA-F]{64}$/;

/** Whether an address clears the shape floor for `chain`. */
function isAddressShaped(address, chain) {
  if (ADDRESS_SHAPED.test(address)) return true;
  return chain === 'ton' && TON_RAW_ADDRESS.test(address);
}
```

`ADDRESS_SHAPED` is untouched and still the only floor for the other 26 chains. The raw form is matched exactly — anchored, fixed 64-hex tail — so it admits the two real workchains and nothing whitespace-shaped. The workchain is `-?\d{1,10}` rather than a literal `0|-1` because the raw format is generic over an int32 workchain; the 64-hex tail is what keeps it tight, and the API does the real per-chain check either way.

The TON error message now names the raw form as well, so a genuinely malformed TON address says what both accepted shapes are.

### Tests — `src/__tests__/api.test.js`

Six added next to the existing batch guards:

- `it.each` over basechain `0:<64 hex>` and masterchain `-1:<64 HEX>` — both sent, `chain: "ton"`, address unchanged (uppercase hex not lowercased);
- `it.each` over three raw-TON lookalikes rejected with 0 fetches — short hex tail (`0:abcdef`), non-numeric workchain (`x:<64 hex>`), and a trailing-whitespace variant (the `--file` guard);
- a raw TON address rejected on `tron`, pinning that the pass is chain-scoped.

The existing friendly-TON test (`UQAbC-…`), the `tron` `--file`-line rejection and the malformed-address cases are all unchanged and still passing.

### Validation

```
npx vitest run src/__tests__/api.test.js -t addressCounterpartiesBatch
  → Tests  29 passed | 214 skipped (243)      (was 23 passed)

npm test
  → Test Files  72 passed (72)
    Tests  2957 passed | 9 skipped (2966)     (was 2951 passed)

npm run lint        → exit 0, no findings
git diff --check    → clean
node -e "JSON.parse(fs.readFileSync('src/schema.json'))" → ok
```

Stubbed-fetch matrix at this head, no network:

```
ton      0:<64 hex>                     -> sent (1 fetch, unchanged)
ton      -1:<64 HEX>                    -> sent (1 fetch, unchanged)
ton      UQAbC-…  (friendly)            -> sent (1 fetch, unchanged)
ton      0:abcdef                       -> rejected (0 fetch)
ton      x:<64 hex>                     -> rejected (0 fetch)
ton      0:<64 hex> + " extra"          -> rejected (0 fetch)
ton      "not an address, just a line"  -> rejected (0 fetch)
tron     0:<64 hex>                     -> rejected (0 fetch)
tron     TQ5NMq… , ethereum 0x…         -> sent (1 fetch)
```

`schema.json` and `SKILL.md` are unchanged — the supported chain list did not move. The changeset gained one clause noting TON takes both forms.

PR #641, opened 2026-09-17
chore: block ticket-ID-shaped tokens in git branch/commit commands

kome12:
## Summary

- nansen-cli is a public repo — internal issue-tracker ticket IDs must never appear in branch names, commit messages, PR titles/bodies, or code/test comments.
- Adds a `PreToolUse` hook (`.claude/settings.json`, gated to git commands via `if: "Bash(git *)"`) that runs `.claude/hooks/no-ticket-refs.sh` before any `git checkout -b`, `git branch -m`, or `git commit`.
- The script scans the command text for a tracker-ID-shaped token (short letter prefix, dash, digits) and blocks the command with an actionable message if found, instead of letting the ticket ID land in the repo.
- Includes an escape hatch (`SKIP_TICKET_CHECK=1` prefix) for genuine false positives — real terms that happen to match the shape (e.g. `SHA-256`).
- Only affects git commands run through Claude Code's Bash tool in this repo; it does not cover manual terminal usage outside Claude Code.

## Test plan

- [x] Pipe-tested the script directly against synthetic hook-input JSON for both block and pass cases (branch names/commit messages with and without ticket-shaped tokens, read-only git commands, the escape hatch)
- [x] Verified live: a real `git checkout -b <ticket-shaped-name>` was blocked with no branch created
- [x] Scanned the script and config for any real internal references before committing


kome12:
Keeping this as a personal local hook instead (.claude/settings.local.json) rather than a shared repo change.

PR #649, opened 2026-09-18
fix(profiler): report compare request failures and match shared tokens by address

kriss39:
## Summary

`research profiler compare` fetches counterparties and balances for both wallets with `.catch(() => null)` and then reads `null` as an empty list. An expired API key, a 429 or a 5xx therefore produces

```json
{"shared_counterparties":[],"shared_tokens":[],"balances":[{"total_usd":0},{"total_usd":0}]}
```

which is indistinguishable from a real "these wallets have nothing in common" answer. For an agent that is a false negative it will act on.

Shared tokens were also intersected on `token_symbol` alone, so two different contracts that share a ticker were reported as a shared holding.

### Changes

- record each failed request instead of swallowing it
- if all four requests fail, rethrow the first error so the command exits with its real code (`UNAUTHORIZED`, `RATE_LIMITED`, ...)
- if some fail, return `incomplete: true`, an `errors` list (`address`, `source`, `code`, `message`) and `null` for the fields that depend on the missing data (`shared_counterparties`, `shared_tokens`, the affected `total_usd`)
- match shared tokens on the token address when both sides report one; fall back to the symbol when either side omits the address (some responses omit it for the native asset). `shared_tokens` is still a list of symbols, or the address when the symbol is empty
- tests: all-fail, one-sided counterparties failure, one-sided balance failure, non-NansenError rejection, same symbol on different contracts, symbol fallback (one side / neither side with an address), empty symbol
- changeset

### Tests

```
 Test Files  75 passed (75)
      Tests  3168 passed | 9 skipped (3177)
```

## Checklist

- [x] Tests pass (`npm test`)


kriss39:
Both points addressed in bac0aa1.

- Asymmetric matching: the address now decides only when both sides report one. If either side has no address for a token, the symbol is used, so the native-asset case (address on one wallet, none on the other) is a shared token again. Two different contracts with the same symbol are still kept apart. Covered by `should fall back to the symbol when only one side reports a token address`.
- Empty symbol: a shared token is labelled by its address when `token_symbol` is empty, never by the internal key. Covered by `should label a shared token by its address when the symbol is empty`.

`npm test`: 75 files, 3167 passed, 9 skipped. Lint and `git diff --check` clean.

kriss39:
Done in 16b1dfc: `code` defaults to `UNKNOWN` for a rejection that is not a `NansenError` (test added), and the changeset, the PR description and a comment next to the match now say what the code does — the symbol is used whenever either side omits the address, the address decides only when both report one.

kriss39:
Done — the all-failed check now compares against the request list (`errors.length === outcomes.length`) instead of a literal 4.

kriss39:
Both done: the all-failed path now rethrows the first recorded failure (whichever request it came from), and the symbol fallback compares symbols case-insensitively while keeping the original spelling in the output. Tests added for a mixed 429/401 failure set and for `USDC`/`usdc`.

kome12 (approved):

PR #651, opened 2026-09-18
fix(api): wait out a server Retry-After instead of capping it at maxDelayMs

kriss39:
## Summary

`calculateBackoff()` clamps a server `Retry-After` to `maxDelayMs` (30s by default). On a 429 with `Retry-After: 60` the client sleeps 30s and retries inside the limiter window, so the remaining attempts are spent on more 429s and the call still fails. `details.retryAfterMs` already reports the real value (60000), only the sleep ignored it.

### Changes

- use the `Retry-After` value as the minimum wait (plus the existing jitter); `maxDelayMs` now only bounds the local exponential backoff
- add `maxRetryAfterMs` (default 120s) to the retry options: if the server asks for more than that, do not retry at all and fail immediately with the 429, which already carries `retryAfterMs`, instead of blocking the CLI for minutes
- tests: a 60s header is waited out in full (nothing fires before 60s, then the retry succeeds), and a 3600s header fails after one attempt with `retryAfterMs: 3600000`
- changeset

Existing behaviour for responses without `Retry-After` is unchanged.

### Tests

```
 Test Files  75 passed (75)
      Tests  3162 passed | 9 skipped (3171)
```

## Checklist

- [x] Tests pass (`npm test`)


kriss39:
Good catch — both checks now test for `null` rather than truthiness, so `Retry-After: 0` (or an HTTP date in the past) means a zero wait plus jitter instead of falling into the exponential backoff. Added a test for the zero case.

kriss39:
Added the `delays.length > 0` guard before the `Math.max` assertion.

kome12 (commented):
Can you address the feedback from [here](https://github.com/nansen-ai/nansen-cli/pull/651#issuecomment-5729216457)? Then the PR should be good go merge!

kriss39:
Done in 55823f7 — the `maxRetryAfterMs` guard is now scoped to 429 (`isRateLimit`). A `Retry-After` on a 5xx is treated as advisory again and capped at `maxDelayMs`, as before this PR. Added a test for a 503 with `Retry-After: 3600` that asserts the retry still happens within 30s, and noted the 5xx behaviour in the changeset.

kriss39:
Widened the margin in 4baaf67: the 5xx advisory-cap test now advances 32 s (30 s cap + up to 1 s jitter), so the assertion no longer sits 1 ms from the maximum delay.

kome12 (approved):

PR #653, opened 2026-09-18
fix(cli): validate --sort direction and field

Bruce039:
## Summary

`parseSort()` (shared by the research commands' `--sort`) only uppercases whatever follows the colon:

```
--sort pnl_usd:sideways   →  order_by: [{ field: "pnl_usd", direction: "SIDEWAYS" }]
--sort :asc               →  order_by: [{ field: "", direction: "ASC" }]
```

Both go to the API as-is and only fail there with a 422, so a typo surfaces late and with a less specific message than the other option validators give.

### Changes

- keep the non-string guard that is now on `main` and, on top of it, reject a direction other than `asc`/`desc` (case-insensitive) with `INVALID_PARAMS` and a message that shows the accepted form
- reject an empty field name the same way
- `field`, `field:asc`, `field:desc` and a trailing colon (`field:`) behave exactly as before; `--order-by` is untouched
- tests updated: the previous "empty field name gracefully" case now asserts the rejection; new cases for bad directions and the trailing-colon default
- changeset

All six `parseSort` call sites use the `field[:asc|desc]` form documented in the README and schema; `trade limit-order list --sort` has its own parser and is not affected.

### Tests

```
 Test Files  75 passed (75)
      Tests  3285 passed | 9 skipped (3294)
```

## Checklist

- [x] Tests pass (`npm test`)
- [ ] `src/schema.json` updated if new commands or flags were added (no new flags)


Bruce039:
Done — the message now quotes the trimmed direction, with a test for the padded input.

kome12 (approved):

PR #662, opened 2026-09-19
fix(schema): sync the documented chain list with the API's published spec

gulshngill:
## Summary

`nansen schema` is the CLI's machine-readable self-description, and its top-level `chains` list is what agents (and shell completion for `--chain`) rely on. It had drifted from what the API accepts in both directions: it still listed `scroll` and `ronin`, which no endpoint serves any more (the API returns 422 for them), and it omitted 20 chains the API does serve.

This PR regenerates the list from the union of every chain enum in the API's published OpenAPI spec (https://api.nansen.ai/openapi.json), sorted and deduplicated (36 chains), and removes the hand-copied duplicates so the list cannot drift again.

- **Removed:** `ronin`, `scroll`
- **Added:** `algorand`, `aptos`, `arc`, `bitcoin`, `bitlayer`, `chiliz`, `citrea`, `gravity`, `hyperliquid`, `injective`, `mantra`, `near`, `robinhood`, `stacks`, `starknet`, `stellar`, `sui`, `ton`, `tron`, `viction`

Not every chain is served by every endpoint; the API keeps validating `--chain` per endpoint, and the `all` wildcard some endpoints accept is deliberately not listed as a chain.

## Changes

- `src/schema.json`: `chains` regenerated from the spec union (sorted, unique).
- `src/cli.js`: the `--help` "Research chains" line is now interpolated from `SCHEMA.chains` instead of a second hand-written copy.
- `README.md`: "Supported Chains" section updated to the same list.
- `src/chain-ids.js`: `EVM_CHAINS` (the CLI's own address-format/ENS gating set) drops `scroll` and `ronin`, so the existing test guard "schema.json chains should be a superset of EVM_CHAINS" keeps holding. `EVM_CHAIN_IDS`, which wallet transfers depend on, is untouched, and no new chains were added to `EVM_CHAINS` (address validation for chains it does not know already falls through to the API).
- `src/api.js`: comment on `COUNTERPARTIES_BATCH_CHAINS` updated to match.
- `src/__tests__/coverage.test.js`: `DOCUMENTED_CHAINS` now reads `SCHEMA.chains` rather than a duplicated literal, plus a new test asserting every entry is a lowercase slug, appears once, and the list is alphabetically sorted.
- `src/__tests__/api.test.js`: the "accept all documented chains" test reads the list from `schema.json`; one test title reworded (`scroll`/`ronin` are no longer "EVM chains this CLI knows").
[...]

PR #663, opened 2026-09-19
fix(cli): keep string option values when they look like JSON keywords

gulshngill:
## Summary

`parseArgs()` runs `JSON.parse()` on every `--option value` pair so that object/array options such as `--filters '{}'` and `--order-by '[...]'` arrive structured. A plain string whose text happens to be a JSON keyword (`true`, `false`, `null`) was therefore silently retyped to a boolean or `null`, with no per-option type awareness. Every string option was exposed, and several handlers had grown their own one-off non-string guards to cope.

`--sort` shows the whole spectrum:

| Command | Before | After |
|---|---|---|
| `nansen research smart-money netflow --chain solana --sort true` | `true` became the boolean `true`; releases before v1.44.2 crashed with `{"success":false,"error":"sortOption.split is not a function","code":"UNKNOWN"}`, current `main` only survives because of a `String()` cast | `order_by: [{ field: "true", direction: "DESC" }]` — the word is a literal field name, exactly like `--sort foo` |
| `... --sort false` / `--sort null` | coerced to `false` / `null`, then silently dropped by the `if (!sortOption)` check — the request went out unsorted with no warning | literal field names `false` / `null`, sent as given |
| `... --sort a --sort b` | joined into the field name `"a,b"` and sent upstream | `{"success":false,"error":"--sort must be \"field\" or \"field:direction\"","code":"INVALID_PARAMS"}` before any request |

Fix is at the shared root, not per consumer: `parseArgs()` keeps the bare keywords `true`/`false`/`null` as strings. Numbers already stayed strings (precision / scientific-notation guard), so this only removes the boolean/null coercion; objects and arrays still parse as JSON. Boolean options (`--premium-labels`, `--neg-risk`, `--apply-blacklist-filter`) already read the strings `'true'`/`'false'` through `resolveBooleanOption()` / `parseBooleanOption()`, so `--flag`, `--flag true`, `--flag false` and an absent flag behave exactly as before.

`parseSort()` gets one small guard: a non-string value (repeated `--sort`, or a JSON object/array) is rejected with `--sort must be "field" or "field:direction"` instead of being stringified into a joined field name. The guard is a single inserted `if` above the existing `String(sortOption).split(':')` line so it merges cleanly with #653, which validates the direction/field on the lines below it.

Side effect worth naming: values like `--query true`, `--question true`, `--api-key null` or `--file true` are now the literal strings a shell user would expect (`--file true` writes to a file named `true`, same as `--file yes`). The existing non-string guards in those handlers stay in place and still fire for JSON objects/arrays and repeated flags.

## Changes

- `src/cli.js` — `parseArgs()` keeps `true`/`false`/`null` option values as strings; comment explains what is and is not coerced.
- `src/query-options.js` — `parseSort()` rejects a non-string value with an actionable `INVALID_PARAMS` error; `parseCsvOption` doc example updated.
- `src/commands/mcp.js`, `src/commands/alerts.js`, `src/wallet.js` — comment-only: the non-string examples no longer cite `--flag true` / `--flag null` since those are strings now.
- `src/__tests__/cli.internal.test.js` — new tests: `true`/`false`/`null` stay strings (single and repeated), `--filters '{}'` / `--order-by '[{"field":"x"}]'` still parse as JSON, numeric values unchanged, `resolveBooleanOption` in all four forms, `parseSort` literal keywords + non-string rejection, and an end-to-end `smart-money netflow --sort true` / repeated `--sort` regression through `parseArgs`.
- `src/__tests__/web-string-input-validation.test.js`, `src/__tests__/mcp-verify.test.js` — the cases that relied on `--flag true` / `--api-key null` arriving as a boolean/`null` now assert the literal-string behaviour, and exercise the non-string guards with a JSON object / repeated flag instead.
- `src/__tests__/alerts.test.js`, `src/__tests__/prediction-market-neg-risk.test.js` — header comments updated to describe the new parser behaviour.
- `.changeset/keep-json-keyword-option-values.md` — patch.

No `src/schema.json` change: no option's documented type changes (`--sort` was already `string`; boolean options still accept `true`/`false`).

## Validation

```
 Test Files  76 passed (76)
      Tests  3212 passed | 2 skipped (3214)
```

`npm run lint` (`eslint .`) passes with no output.

Live check with the API base pointed at a closed local port (no network): `--sort true`, `--sort false` and `--sort null` now reach the request stage (previously `false`/`null` were dropped before it), repeated `--sort a --sort b` returns the `INVALID_PARAMS` error above, and `parseArgs(['--filters','{}','--order-by','[{"field":"x"}]','--sort','true','--limit','10'])` yields `{"filters":{},"order-by":[{"field":"x"}],"sort":"true","limit":"10"}`.

## Checklist

- [x] Tests pass (`npm test`)
- [x] `src/schema.json` updated if new commands or flags were added (n/a — no new flags, no type changes)
- [x] `README.md` updated if new top-level commands or categories were added (n/a)


gulshngill:
**Merge position: 4 of 9.** Merge after #662.

Changes the shared argument parser, so it belongs ahead of the PRs that add options on top of it (#667, #668). Shares `src/wallet.js` with #666 and `src/cli.js` with #666, #667, #668 and #669.

Note #669 adds its own `resolveBooleanOption` tests in a separate file. Both sets pass either way, so the only cost is two describe blocks covering the same function.

gulshngill:
Addressed the reviewer note in 5a7781d. The boolean branch is reachable for bare flags because `parseBooleanOption()` reads from both `options` and `flags`: `--apply-blacklist-filter` arrives as `flags["apply-blacklist-filter"] === true`, while an explicit value arrives through `options` as a string. Added an explanatory comment and regression coverage for the bare-flag path.

gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
Addressed the remaining low-severity review finding in 4a2abe5: after the non-string guard, `parseSort()` now calls `sortOption.split(":")` directly. The focused 601-test suite and lint both pass.

gulshngill:
@nansen-pr-reviewer re-review

PR #664, opened 2026-09-19
fix(trade): screen addresses for sanctions before quoting and executing swaps

gulshngill:
## Summary

`nansen trade quote` and `nansen trade execute` now run the same pre-trade compliance screening that `nansen bridge` and `nansen perp` already perform, so DEX swaps are no longer the one trading surface in the CLI without it — for same-chain and cross-chain (`--to-chain`) swaps alike.

Both commands reuse the existing fail-closed `screenOrThrow()` helper (`POST /api/v1/sanctions/screen`), exactly as `bridge execute` does:

- **`trade quote`** screens the signing wallet plus any distinct destination wallet (`--to-wallet`, or the destination auto-derived for a cross-chain swap) *before* the quote request goes out, so a flagged wallet never receives a signable quote.
- **`trade execute`** screens once, immediately before the per-quote signing loop — covering the resolved signing wallet, the wallet the quote was built for, and any distinct recipient. A quote stays valid for up to an hour and the signed transaction is broadcast with no further check, so this is the last gate before funds move. Screening outside the loop means a refusal is final rather than "try the next quote".

A flagged address aborts with code `SANCTIONED`; a screening call that cannot complete (or a response that doesn't cover every requested address) aborts with `SCREENING_UNAVAILABLE`. In both cases nothing is signed or broadcast and the quote is left unconsumed, so a fresh attempt is still possible once the situation is resolved.

Two smaller behaviour notes:

- A local wallet file with no key for the quote's chain now fails with the same actionable "cannot confirm the quote was built for this wallet" message the signing branch gives, instead of proceeding to a screen it cannot satisfy.
- A quote that records no wallet at all (only reachable from a hand-edited or truncated quote file) is refused with `SCREENING_UNAVAILABLE` and a prompt to re-quote, rather than silently skipping the check.

Screening requires API access — the same requirement `bridge` and `perp` already have — so an unauthenticated call fails closed rather than trading unscreened.

No new commands or flags, so `src/schema.json` is unchanged.

## Changes

- `src/trading.js` — import `screenOrThrow`; new exported `tradeScreeningAddresses()` helper (drops blanks, de-duplicates EVM addresses case-insensitively while keeping base58 addresses distinct by exact spelling); screen in the `quote` handler before the quote request; resolve the signer ahead of the signing loop and screen in the `execute` handler before it.
- `src/__tests__/trading-sanctions-screening.test.js` — new: the gate for both commands across local, WalletConnect and provider-signed paths, plus the address-list helper.
- `src/__tests__/trading.test.js`, `src/__tests__/trading-quote-reuse.test.js`, `src/__tests__/trade-validation.test.js`, `src/__tests__/blank-option-values.test.js` — existing trade tests pass an API double whose screen reports clean (they previously passed `null`, since these handlers made no API-instance calls).
- `skills/nansen-trading/SKILL.md` — document the screen and its two error codes under Notes; note that the API key is needed for it.
- `.changeset/trade-sanctions-screening.md` — patch.

## Validation

```
npm test
```
```
Test Files  77 passed (77)
     Tests  3215 passed | 2 skipped (3217)
  Duration  144.72s
```

```
npm run lint
```
Clean, no output.

All network calls in the new tests are mocked (API instance double + stubbed `fetch` for the trading backend and RPC); no real trade is quoted or executed. The tests assert that the screen call precedes the `/quote` and `/execute` calls, that a flagged signer or destination aborts with no transaction sent and the quote left unconsumed, and that a screening failure or an incomplete screening response aborts the same way.

## Checklist

- [x] Tests pass (`npm test`)
- [x] `src/schema.json` updated if new commands or flags were added — n/a, no new commands or flags
- [x] `README.md` updated if new top-level commands or categories were added — n/a, no new commands


gulshngill:
**Merge position: 6 of 9.** Merge before #667.

Both this and #667 change `src/trading.js` and `skills/nansen-trading/SKILL.md`. This one inserts the pre-trade check in the quote and execute paths; #667 adds a confirmation gate slightly later in the same execute path. Landing screening first means #667 rebases onto the final ordering rather than the reverse.

kome12 (approved):

PR #665, opened 2026-09-19
ci: make the changeset presence check actually run on pull requests

gulshngill:
## Summary

The changeset-presence check has never run on a pull request. `pretest` runs `scripts/check-changeset.js`, which diffed the working tree against a **local** `main` ref. The `test` job checks out with `actions/checkout` at its default depth — one commit, one ref, detached HEAD — so `main` does not exist locally, `git diff main ...` failed with `fatal: bad revision 'main'`, and the script's bare `catch` swallowed it and exited 0. Neither signal the script exists for could fire: not the "No new changeset file found" warning, and not the hard failure for a changeset whose frontmatter names the wrong package (which silently skips the version bump at release time).

You can see the swallowed error in the log of any recent PR run, immediately after `> node scripts/check-changeset.js`.

This PR fixes both halves: the script now resolves its comparison base explicitly and says when it cannot, and the `test` job fetches the tip of `main` so there is a base to resolve.

I picked the single `git fetch --no-tags --depth=1 origin +refs/heads/main:refs/remotes/origin/main` over `fetch-depth: 0` on the checkout (what `publish` uses). The check only compares two trees, so it needs the tip commit, not history — fetching one commit is cheaper than a full clone, on every matrix entry of every PR.

## Changes

`scripts/check-changeset.js`
- Resolve the base as `origin/main`, falling back to a local `main`. Preferring the remote-tracking ref also avoids comparing against a stale local `main` in a developer checkout.
- If neither ref resolves, print `[changeset] Could not resolve a base branch (tried origin/main, main); skipping check.` and exit 0 instead of exiting silently.
- Keep the catch-all, but have it log a one-line notice (`[changeset] Skipping check: <reason>`) rather than swallow. It still never fails the test run for reasons unrelated to changeset files.
- Detect "we are on the base" by ancestry (`git merge-base --is-ancestor HEAD <base>`) as well as by branch name, so a push to `main` — checked out in CI as a detached HEAD at that commit — still no-ops.
- Split the body into small exported functions (`resolveBaseRef`, `isOnBase`, `findNewChangesets`, `findInvalidChangesets`, `run`) so it can be unit-tested. The `#!/usr/bin/env node` entry point is unchanged: run directly it behaves exactly as before, printing to stderr and exiting 0/1. Argument arrays now go through `execFileSync` instead of a shell string.

`.github/workflows/ci.yml`
- Add a `git fetch --no-tags --depth=1` step to the `test` job, after checkout. No other job changed; all action SHAs untouched.

`src/__tests__/check-changeset.test.js` (new)
- 13 tests over throwaway git repositories built in a temp dir, including depth-1 single-branch detached clones that reproduce the CI checkout. All clones are local `file://` — no network, so `npm test` stays offline.

Behaviour that is deliberately unchanged: a missing changeset warns and exits 0; a new changeset whose frontmatter lacks `"nansen-cli":` exits 1.

## Validation

```
$ npm test
 Test Files  77 passed (77)
      Tests  3211 passed | 2 skipped (3213)
   Duration  141.02s

$ npm run lint
(no output, exit 0)
```

Simulated CI checkout — a depth-1, single-branch, detached clone of this branch, which is what the `test` job gets:

```
$ git clone --depth 1 --single-branch --branch fix/ci-changeset-check-base <repo> ci-checkout
$ cd ci-checkout && git checkout --detach

=== BEFORE (script as on main) ===
fatal: bad revision 'main'
exit=0                                  # nothing from the check itself

=== AFTER (this branch), base not fetched ===
[changeset] Could not resolve a base branch (tried origin/main, main); skipping check.
exit=0

=== AFTER, with the new fetch step ===
[changeset] No new changeset file found on this branch. If this PR changes user-facing behavior, add one: npx changeset
exit=0

=== AFTER, with a changeset added ===
exit=0                                  # silent, as intended

=== AFTER, with a changeset naming the wrong package ===
[changeset] ERROR: .changeset/example.md has incorrect or missing package reference.
  Expected: "nansen-cli": <patch|minor|major>
  An invalid package name will silently skip the version bump during release.
exit=1
```

This PR is its own end-to-end test: it adds no changeset, so the `test` job should now print the yellow warning.

## Checklist

- [x] Tests pass (`npm test`) — output above
- [x] `npm run lint` passes
- [x] `src/schema.json` updated if new commands or flags were added — n/a, no CLI surface change
- [x] `README.md` updated if new top-level commands or categories were added — n/a
- [x] Changeset added — **not needed: tooling/CI only, nothing user-facing in the published package**


gulshngill:
**Merge position: 1 of 9.** No file overlap with any other open PR in this batch, and landing it first means every PR after it actually gets the changeset check this one repairs.

Blocked only by the review-policy rule on workflow files, which needs a human override rather than a code change.

PR #667, opened 2026-09-19
feat(trade): add --dry-run and TTY confirmation (--yes) to trade and bridge execute

gulshngill:
## Summary

`nansen trade execute` and `nansen bridge execute` move funds irreversibly, but until now there was no way to preview what they would send and no acknowledgement step before broadcasting. This PR adds both, to each command.

**`--dry-run`** runs the same validation a real execute runs, prints the trade or transfer that *would* be sent, and stops before anything is signed:

```
  Trade plan — Base
    Quote:          1708900000000-abc123 via lifi (quote 1 of 2)
    Sell:           1000000 base units of 0x833589fc…
    Buy:            ~440000000000000 base units of 0xeeee…
    Max spend:      1000000 base units
    Wallet:         0xAbC0…0001
    Trading fee:    $0.01
    Network fee:    $0.02
    Approval:       required → 0x1231deb6… (scoped to 1000000 base units)
    Outcome check:  on, before broadcast

  DRY RUN — nothing was broadcast. No transaction was signed and the quote was not consumed.
```

The gate sits before any wallet material is loaded, so a dry run needs no wallet password, cannot sign anything, and leaves the quote reusable. On EVM it also reads the current token allowance and runs the pre-broadcast revert simulation when no approval is outstanding. On a multi-step bridge transfer it stops before the first step.

**Confirmation.** When stdin is an interactive terminal, both commands now print that plan and ask `Broadcast this transaction? [y/N]` before broadcasting. Anything but `y`/`yes` aborts with nothing signed. `--yes` (`-y`), or `NANSEN_YES=1`, skips the question.

| stdin | `--yes` / `NANSEN_YES=1` | Behaviour |
|---|---|---|
| Terminal | no | Prints the plan, asks `Broadcast this transaction? [y/N]`; declining aborts |
| Terminal | yes | Broadcasts, no prompt |
| Not a terminal (agents, CI, pipes) | no | **Unchanged** — broadcasts without prompting |
| Not a terminal | yes | Unchanged — broadcasts without prompting (`--yes` is a no-op) |

Non-interactive behaviour is deliberately identical to today's, so existing automation keeps working untouched and `--yes` is always safe to pass. The plan and prompt are only emitted in dry-run mode or immediately before an interactive prompt (the prompt itself goes to stderr), so piped `--pretty` / `--format` output is unaffected.

**Exit codes** (`trade execute`, `bridge execute`):

| Code | Meaning |
|---|---|
| `0` | Broadcast succeeded, or the dry run completed |
[...]

PR #666, opened 2026-09-19
docs(cli): list every wallet subcommand in --help, README and schema

gulshngill:
## Summary

Every `nansen wallet` subcommand is now discoverable from the three places a user or an agent looks first: `nansen --help`, `README.md` and `nansen schema`.

The help banner listed seven of the nine wired wallet subcommands — `send` and `secure` were missing, so nothing in `--help` revealed that the CLI can send tokens or migrate a saved wallet password into the OS keychain. README omitted `secure`; `nansen schema wallet` omitted both `secure` and `forget-password`.

To stop this drifting again, the banner's wallet line is now generated from a single exported list that lives next to the dispatcher, and a new drift-guard test pins the banner, README and `src/schema.json` to the real command surface — for every command group, not just `wallet`. That test also caught the `perp` line, which named five of its ten subcommands; it is corrected here as a one-line list fix.

No behaviour change: help text, docs and schema output only.

## Changes

- `src/wallet.js` — export `WALLET_SUBCOMMANDS`, the ordered list of subcommands the dispatcher accepts (`help` excluded: it is the fallback, not a capability).
- `src/cli.js` — the `wallet` line of the help banner is built from that list, so it cannot fall behind the dispatcher. The `perp` line now names all ten of its subcommands (`transfer`, `approve-builder-fee`, `orders`, `account`, `meta` were missing).
- `README.md` — the wallet list adds `forget-password` and `secure`; the Wallet section shows both commands in its example block.
- `src/schema.json` — adds `wallet forget-password` and `wallet secure` with descriptions, and gives every wallet subcommand at least one example.
- `src/__tests__/command-surface.test.js` — new drift guard: every group in the schema has a banner line; every subcommand of a group is named on that line (`research` is exempt as a curated catalogue, documented in the test); no banner line advertises a subcommand that does not exist; and the wallet set matches across the dispatcher source, the banner, README, `nansen wallet help` and the schema.
- `.changeset/wallet-subcommand-discoverability.md` — patch.

## Validation

```
npm test
 Test Files  77 passed (77)
      Tests  3207 passed | 2 skipped (3209)
```

```
npm run lint
(eslint . — no findings)
```

Drift guard verified by mutation: removing `secure` from the exported list and reverting the `perp` line makes 4 of the 9 new assertions fail with the exact missing names.

## Checklist

- [x] Tests pass (`npm test`)
- [x] `src/schema.json` updated if new commands or flags were added
- [x] `README.md` updated if new top-level commands or categories were added
[...]

PR #668, opened 2026-09-19
feat(cli): add --debug request tracing with credential redaction

gulshngill:
## Summary

Adds a general request trace, so auth, latency, retry and response-shape failures can be diagnosed without guesswork.

`--debug` (a global flag) or `NANSEN_DEBUG=1` prints one line per HTTP request to **stderr** — method, URL, response status, latency in ms, retry decisions (attempt, reason, delay) and the server request id when the response carries one. stdout still carries only the JSON/CSV payload, so piping into `jq` or a CSV reader is unaffected:

```bash
nansen research token screener --chain solana --debug
nansen research token screener --chain solana 2>trace.log | jq .   # trace to a file, JSON to jq
```

```
[nansen:debug] http.request method=POST url=https://api.nansen.ai/api/v1/token-screener attempt=1/4
[nansen:debug] http.response method=POST url=https://api.nansen.ai/api/v1/token-screener status=429 duration_ms=182 request_id=6f1c0f2a-0000-4000-8000-0000000000aa attempt=1
[nansen:debug] http.retry method=POST url=https://api.nansen.ai/api/v1/token-screener status=429 attempt=1 reason=retry-after delay_ms=1100 retry_after_ms=1000
[nansen:debug] http.request method=POST url=https://api.nansen.ai/api/v1/holders?apikey=[redacted]&token_address=[redacted]&chain=solana attempt=2/4
[nansen:debug] http.response method=POST url=https://api.nansen.ai/api/v1/token-screener status=200 duration_ms=143 request_id=6f1c0f2a-0000-4000-8000-0000000000ab attempt=2
```

(Everything above is fake — no real key, id or host response.)

**Redaction guarantee.** The trace never prints credentials and never prints request or response bodies. Concretely:

- Header values are not traced at all — not `apikey`, not `Authorization`, not `Payment-Signature`, not even the harmless ones. One rule, no exceptions to get wrong.
- Query-string values are blanked whenever the parameter name mentions key, token, secret, signature, password or auth — `token_address` is blanked too, which is over-eager on purpose.
- Any remaining value shaped like a credential is blanked wherever it appears: private keys (32 bytes of hex), BIP-39 mnemonics, JWTs, long base58 secrets, long opaque payment payloads, and inline `Bearer …` / `Payment …` credentials.
- URL userinfo (`https://user:pass@…`) is dropped, and long values are truncated.
- Request ids, addresses, chains and paths stay readable — a trace that hides those is not worth pasting into a bug report.

`redact()` is a standalone helper in `src/debug.js` with its own tests, and each transport path has a test asserting that a realistic-but-fake key never reaches the output.

**What already existed and what was unified.** There was no general trace. The only debug machinery in the repo was two x402 wallet-lookup notes in `src/privy.js` and one in `src/api.js`, each gated on a bare `DEBUG` env var and printed via `console.error` — narrow and undocumented. Those now go through the same helper, so there is one switch, one output stream and one redaction pass rather than a parallel mechanism; a bare `DEBUG` keeps working as an alias for `NANSEN_DEBUG`. `--verbose` already parsed as a flag but did nothing, and still does nothing — this PR deliberately does not claim it.

**Coverage.** The shared API client in `src/api.js` is traced end to end: the main request loop (every attempt, network failures, retry decisions) and the x402 paid retry, plus a `http.cache_hit` line so a locally cached answer does not look like a silent request. Not covered, and better as follow-ups than as scope creep here: the AI-agent SSE stream, trading/bridge/limit-order calls, direct RPC node calls, and the telemetry and update-check background fetches — each has its own fetch and its own error handling.

## Changes

- `src/debug.js` (new) — the switch (`--debug` / `NANSEN_DEBUG` / `DEBUG`), the stderr writer, `redact()` / `redactUrl()` / `redactHeaders()` / `truncate()`, and the `http.*` trace helpers.
- `src/api.js` — trace each attempt of the request loop (request, response, retry decision, transport error), the cache hit, and the x402 paid retry; one local helper reads the request id off the response header.
- `src/cli.js` — `--debug` turns tracing on before anything can make a request; `--debug` documented in the help banner.
- `src/privy.js`, `src/api.js` — the three pre-existing `DEBUG`-gated notes now route through the shared helper.
- `src/schema.json` — `debug` added to `globalOptions` (shell completions pick it up from there).
- `README.md` — a short "Debugging" section with a sample trace and the explicit "never includes credentials or response bodies" note, plus a row in Key Options.
- `.changeset/debug-request-tracing.md` — minor.
- Tests — `src/__tests__/debug.test.js` (redaction and the switch) and `src/__tests__/api-debug-trace.test.js` (both transport paths, retry, cache hit, flag wiring). All network calls are mocked.

**Expected rebase points.** #651 (Retry-After) edits the same retry branch in `request()` that the retry trace hooks into — the trace line sits after the delay is computed and reports whatever that PR decides, so the two agree on behaviour; expect a context conflict, not a semantic one. #628, #661 and #596 touch `src/api.js` elsewhere and should not overlap.

## Validation

```
npm test
 Test Files  78 passed (78)
      Tests  3220 passed | 2 skipped (3222)

npm run lint
> eslint .   (clean, exit 0)
```

Also exercised by hand against an unreachable local port, confirming the trace goes to stderr while stdout stays pure JSON, and that `--debug` and `NANSEN_DEBUG=1` behave identically.

## Checklist

- [x] Tests pass (`npm test`)
- [x] `src/schema.json` updated if new commands or flags were added
- [x] `README.md` updated if new top-level commands or categories were added


gulshngill:
Both findings addressed in 8fba5e7.

**`--debug` missing from `VALUELESS_FLAGS` (high)** — confirmed and fixed. Reproduced first: `nansen --debug research token screener --chain solana` parsed `research` as the flag's value, so tracing never turned on and the run fell through to the deprecated `nansen token screener` alias instead of the command that was typed. Added `'debug'` to the set, plus a regression test asserting `parseArgs(['--debug','auth','status'])` yields `flags.debug === true` with `_` intact, and that `runCLI` enables tracing when the flag comes first. The test fails on the unfixed parser.

**`INLINE_SCHEME` matching prose (medium)** — took the stricter half of your suggestion: dropped `Payment` and `Signature` from the keyword list and required the credential to be 8+ characters, so the pattern is now `/\b(Bearer|Basic)\s+[\w.~+/=-]{8,}/i`. A real payment signature is still covered by header-name matching and by the value-shape rules, which is why the keywords were redundant rather than protective. Added tests pinning that `payment outcome unknown`, `signed payment was transmitted` and `Signature verification failed for header` pass through unchanged, that `Bearer token` is left alone, and that a genuine `Bearer eyJ...` is still redacted.

`npm test` — 78 files, 3224 passed, 2 skipped. `npm run lint` clean.

gulshngill:
**Merge position: 8 of 9.** Merge after #661, #662 and #666.

Shares `src/api.js` and `README.md` with #661 and #662, and `src/cli.js` plus `src/schema.json` with #662, #666 and #667.

One live conflict to expect: #651 edits the same retry branch this PR's retry trace hooks into. The trace reports whatever that PR decides, so it is a context conflict rather than a behavioural one.

gulshngill:
@codex review

gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
@nansen-pr-reviewer re-review

gulshngill:
@nansen-pr-reviewer re-review

PR #669, opened 2026-09-19
test(cli): cover boolean option parsing, subcommand help and changelog filtering

gulshngill:
## Summary

Three exported helpers shape most of the text an agent reads before it calls anything, and none of them had direct assertions on what they actually produce:

- `resolveBooleanOption()` — the `--flag` / `--flag true` / `--flag false` / absent tri-state used by every boolean option. Zero direct tests.
- `generateSubcommandHelp()` — the body of every `nansen <cmd> <sub> --help`. Only ever asserted via two `Example:` substrings, so the params list, the required marker, the credit-cost line and the returns line were unpinned.
- the `changelog --since` filter — covered for its two error paths, but nothing asserted *which* entries survive a filter.

This adds a new test file, `src/__tests__/help-ux.test.js`, that pins the shape of that output. It is deliberately a new file rather than an edit to `cli.internal.test.js`, which several open PRs are already touching.

Writing the help tests turned up a bug, fixed here (see Findings).

## Changes

- **`src/__tests__/help-ux.test.js`** (new, 25 tests + 1 skipped):
  - `resolveBooleanOption`: bare `--flag`; explicit `true` / `false` / `1` / `0` in any case; absent (and not leaking from a neighbouring boolean option); and that an unrecognised value never resolves to `true`. The explicit-value cases assert both the string and JSON-boolean spellings of the value, so they hold however the option value happens to be parsed.
  - `generateSubcommandHelp`: the first line (name + description); the params line with its `*` required marker, `(default)` and `[enum|values]`; the credit-cost line — present with the cached per-tier cost, pluralised correctly for a single credit, and omitted entirely when no cost is cached; the returns line; and both example forms (a hand-written schema example used verbatim vs. one generated with every required option appended). Costs are read from a seeded cache under a temp `HOME`, and the two `runCLI --help` tests stub `fetch` and assert it is never called.
  - `changelog --since`: run against a synthetic changelog so the expectations are exact — the boundary version is included (the filter is `>=`), everything older is dropped including its body lines, `1.10.0` survives a `--since 1.9.0` (numeric, not string, ordering), a `major.minor` value behaves like its `.0` form, no `--since` prints the file unchanged, and a pre-release value is refused with the actionable message instead of being guessed at.
- **`src/cli.js`**: help lookup now checks both the top-level command and the research category of the same name (top-level wins when both have the subcommand), and points the generated example at the `research` path when the subcommand came from there.
- **`.changeset/research-perp-subcommand-help.md`**: patch, for that fix only.

`compareSemver` already has its own `describe()` block in `src/__tests__/semver.test.js` (its canonical home since it moved to `src/semver.js`), so it is not re-tested here — that file explicitly asks for the comparison arithmetic to live there and the command behaviour to live with the command.

## Findings

1. **Bug, fixed: `nansen research perp screener --help` (and `leaderboard`) never showed the subcommand's help.** `perp` is both the top-level trading command and a research category, with different subcommands under each. The lookup resolved the *name* to the trading command, found no `screener` under it, and returned `null`, so `--help` fell back to the category listing — which ends with `Use: nansen research perp <subcommand> --help`, i.e. the exact command the caller had just run. The params, credit cost and example were unreachable through help at all. Three tests in this PR fail without the fix.

2. **Not fixed, deliberately: `resolveBooleanOption` silently ignores anything that is not `true`/`false`/`1`/`0`.** `--premium-labels yes`, `--premium-labels ""` (an unset shell variable) and a repeated `--premium-labels true --premium-labels false` all resolve to `undefined`, which callers read as "not supplied" and fall back to the server default. The caller gets no error and no hint that their value was dropped. That is at odds with the direction of recent work rejecting blank option values elsewhere, but making it throw changes behaviour across every command that takes a boolean option, so it is recorded as one skipped test with the correct expectation rather than changed here. An active test does pin the part that must never regress: an unrecognised value must not resolve to `true`, i.e. must never silently opt a caller *into* a premium switch.

3. **Not fixed, out of scope: generated examples are not runnable for ~35 subcommands.** The example generator only knows placeholder values for `address`, `token`, `query`, `symbol` and `date`, so everything else renders as `<val>` — e.g. `nansen perp order --coin <val> --side <val> --size <val> --price <val>`. For options that carry an `enum` or a `default` in the schema (`--side`, `--type`, `--direction`), the example could use a real value the params line already prints two lines above.

4. **Not fixed, out of scope: required positional arguments never appear in the params list.** `generateSubcommandHelp` renders `options` only, so `mcp install` / `mcp uninstall` show `Params (* required): --dry-run` and never mention the required `client` positional or its allowed values. Today both describe it in their prose description, so nothing is actually undiscoverable — but a future subcommand with a positional would silently lose it from help.

5. **Minor, reported only: version comparison ignores pre-release tags.** `1.45.0-beta.1` compares equal to `1.45.0`, because the parser reads each dot-separated component with `Number` and a `NaN` component falls back to `0`. `changelog --since` is unaffected — it rejects a pre-release value outright, which this PR pins as a test — but the update notifier would treat a pre-release `latest` as the same version as the release.

## Validation

```
npm test
 Test Files  77 passed (77)
      Tests  3223 passed | 3 skipped (3226)

npm run lint
(no output, exit 0)
```

The skipped count includes the one skipped test added here; the other two are pre-existing.

## Checklist

- [x] Tests pass (`npm test`)
- [x] `src/schema.json` updated if new commands or flags were added — n/a, no new commands or options
- [x] `README.md` updated if new top-level commands or categories were added — n/a
- [x] Changeset added (`.changeset/`) — added for the help-lookup fix only; the rest of this PR is tests


gulshngill:
**Merge position: 9 of 9.** Merge last.

The tests here pin the generated help output, which #666, #667 and #668 all change. Rebasing this onto the finished state avoids re-pinning the assertions more than once. The `src/cli.js` change itself is a four-line fix in `generateSubcommandHelp` that none of the others touch.

gulshngill:
**Correction to the position above: this is now 10 of 10**, after #670, which registers the `cache` command in `src/schema.json` and adds its help text. Still last, and for the same reason: the assertions here pin generated help output that the PRs ahead of it change.

gulshngill:
Addressed both review findings in 23f649a. Test-only; `src/cli.js` and the changeset are unchanged.

## MEDIUM — `vi.resetModules()` in `afterEach` — confirmed, fixed

The finding holds. The nested `credit cost line and --help rendering` block reset the module registry in both `beforeEach` (line 171) and `afterEach` (line 178). The `afterEach` call is redundant, since the `beforeEach` reset runs immediately before the next test anyway, and it left the registry cleared for whatever ran next. `src/__tests__/cost-cache.test.js` resets only in `beforeEach` and its `afterEach` just restores env and removes the temp dir, so this now matches. `afterEach` keeps the `HOME` restore and the temp-dir cleanup.

I also stress-tested the ordering concern rather than taking it on faith:

```
--- shuffle-only run 1..4 (help-ux + cost-cache + cli) ---
 Test Files  3 passed (3)
      Tests  66 passed | 1 skipped (67)
```

Stable across four randomised runs. Worth recording one side-note for anyone who tries the same: under `--sequence.concurrent` three of these cases fail — but they fail identically **before** this change, and the untouched `cost-cache.test.js` fails the same way (9 cases), because forcing concurrency breaks the `HOME`-swapping pattern both files use. That is an artifact of the flag, not of the `afterEach` reset, and the suite does not run concurrent.

## LOW — stub ordering in `runHelp` — premise correct, conclusion inverted; left alone

The described ordering is accurate: `vi.resetModules()` runs in `beforeEach`, then `runHelp` installs the `fetch` stub, then `await import('../cli.js')`. But that is the *safe* ordering, not a fragile one — the spy is in place **before** the dynamic import, so an import-time request would be recorded, not missed.

I tested it instead of arguing it. Temporarily adding a top-level `await fetch('https://example.invalid/import-time-probe')` to a module in the `cli.js` graph:

```
 × prints the deprecation note above the generated body for a moved command
 × answers a research subcommand help request with that subcommand, not with the category listing
      "https://example.invalid/import-time-probe",
    232|         expect(fetchSpy).not.toHaveBeenCalled();
 Tests  2 failed | 23 passed | 1 skipped (26)
```

Both `runHelp` cases failed loudly and named the offending URL. So the stated failure mode — "the spy would silently miss the call" — does not occur, and moving the stub earlier would change nothing (`vi.resetModules()` lives in `beforeEach`, which already runs before the stub *and* before the import). I reverted the probe; the module is byte-identical.

I did take the finding's second suggestion, which is the accurate framing, as a comment-only change: the seeded caches are the real network guard and the spy is belt and braces. The old comment named only `update-check.json` and missed the more important one — the `--help` path awaits an inline spec fetch, so it is the seeded **cost map** that keeps `runHelp` offline. Demonstrated by running the real binary with the cost cache cold, under a preload that throws on any outbound request:

```
=== COLD cost-map cache, fresh update-check ===
!!! NETGUARD: fetch https://api.nansen.ai/openapi.json
perp screener — Screen perpetual futures contracts
...
=== same, with a FRESH seeded cost-map ===
perp screener — Screen perpetual futures contracts
...
Cost: 1 credit (Free tier) / 1 credit (Pro tier)
```

The one genuinely unguarded load is the static `import ... from '../cli.js'` at the top of the file, which executes the graph before any stub exists — but that is inherent to importing the module under test, applies to every test file in the repo, and neither suggested fix addresses it. Left alone.

## Local verification of the `src/cli.js` fix

Ran the real binary. Before (`src/cli.js` at `origin/main`):

```
$ node src/index.js research perp screener --help
research perp — Perpetual futures analytics
Subcommands: screener, leaderboard
Use: nansen research perp <subcommand> --help

$ node src/index.js research perp leaderboard --help
research perp — Perpetual futures analytics
Subcommands: screener, leaderboard
Use: nansen research perp <subcommand> --help
```

Both answer a subcommand help request by telling you to run the command you just ran. After:

```
$ node src/index.js research perp screener --help
perp screener — Screen perpetual futures contracts
Params (* required): --days (30), --trader-type [all|sm|whale|public_figure|high_winrate_hl_perps_trader], --sectors-filter, --sm-label-filter, --trader-label-filter
Cost: 1 credit (Free tier) / 1 credit (Pro tier)
Example: nansen research perp screener

$ node src/index.js research perp leaderboard --help
perp leaderboard — Perpetual futures PnL leaderboard
Params (* required): --days (30), --premium-labels
Cost: 5 credits (Free tier) / 5 credits (Pro tier)
Example: nansen research perp leaderboard
```

No regression on the top-level path or elsewhere — the trading command keeps its own subcommands, and the required marker, defaults, enums, deprecation notes and hand-written examples all still render:

```
$ node src/index.js perp order --help
perp order — Place a perp order (limit or market, with optional take-profit/stop-loss)
Params (* required): --coin*, --side* [buy|long|sell|short], --size*, --price*, --type (limit) [limit|market], --tif (Gtc) [Gtc|Ioc|Alo], --slippage (0.03), --take-profit, --stop-loss, --wallet
Example: nansen perp order --coin <val> --side <val> --size <val> --price <val>

$ node src/index.js perp positions --help
perp positions — View open positions
Params (* required): --wallet
Cost: 0 credits (Free tier) / 0 credits (Pro tier)
Example: nansen perp positions

$ node src/index.js research token screener --help
token screener — Discover and filter tokens
Params (* required): --timeframe (24h), --chain (solana), --include-stablecoins (true)
Cost: 1 credit (Free tier) / 1 credit (Pro tier)
Example: nansen research token screener --chain solana

$ node src/index.js wallet create --help
wallet create — Create a new wallet. Explicitly blank --name values are rejected for local and Privy wallets.
Params (* required): --unsafe-no-password
Example: nansen wallet create

$ node src/index.js token holders --help
Note: "nansen token" is deprecated. Use "nansen research token" instead.

token holders — Token holder analysis
Params (* required): --chain (solana), --token*, --premium-labels
Cost: 5 credits (Free tier) / 5 credits (Pro tier)
Example: nansen research token holders --token 0x... --chain solana
```

`nansen perp screener --help` now also resolves, with its example pointing at the `research` path — intended, and called out in the changeset. `trade quote --help` still shows the handler's own rich usage (it is skipped by the schema lookup). All of the above re-run under the throw-on-network preload with a warm cache: zero requests, zero spawns, no credits spent. Nothing that touches the API or trades was run.

## Suite

```
$ npm run lint
> eslint .
[exit 0]

$ npm test
 Test Files  77 passed (77)
      Tests  3223 passed | 3 skipped (3226)
```

Identical to the pre-change baseline. `package-lock.json` untouched.

PR #671, opened 2026-09-20
fix(api): keep non-JSON error bodies and retry non-JSON 429s

0xShadowX:
## Summary

Two problems in the same branch of `NansenAPI.request()`, the one that handles a response whose body is not JSON (gateway HTML, plain text from a proxy or rate limiter).

**1. The body is lost.** `response.json()` is tried first and, when it fails, `response.text()` is called on the same response. A fetch body can only be read once, and `json()` consumes it even when the parse fails, so on a real `Response` the second read throws `Body is unusable` and the error goes out with `body: null`. The existing test for this path uses a `{ json, text }` mock, which allows the double read and so does not see it.

```js
const r = new Response('<html>502</html>', { status: 502 });
await r.json().catch(() => r.text());   // TypeError: Body is unusable: Body has already been read
```

**2. A non-JSON 429 is not retried.** This branch retries only on `status >= 500`, while the JSON branch uses `retryOnStatus` (429, 500, 502, 503, 504). A rate limiter that answers in plain text therefore fails on the first attempt with `maxRetries` untouched, and its `Retry-After` is never read.

### Changes

- read a body that the server declares as non-JSON (`content-type` present and not JSON) as text once; it is still parsed as JSON when it turns out to be JSON, otherwise the text is kept for the error details. Responses declared as JSON, and responses without a content type, go through `response.json()` exactly as before, with the previous `text()` fallback kept
- the non-JSON branch now uses the same `retryOnStatus` policy as the JSON branch, records `retryAfterMs` in the error details, passes it to the backoff, and maps the status through `statusToErrorCode` so a plain-text 429 is `RATE_LIMITED` like a JSON one
- `stringHeader` from `response-meta.js` is exported and reused for the two header reads
- tests with real `Response` objects: the 502 HTML body survives in `details.body`; a plain-text 429 with `Retry-After: 2` is retried after ≥2s and succeeds and carries `RATE_LIMITED`; a plain-text 400 is not retried and keeps its body; JSON sent with a non-JSON content type still parses; a body with no content type at all still goes through `json()` and is reported with `body: null` as before (no crash)
- changeset

Checked against a local HTTP server that answers two plain-text 429s and then JSON: the CLI now returns the data with `_meta.retriedAttempts: 2`; before it failed on the first 429.

### Tests

```
 Test Files  78 passed (78)
      Tests  3289 passed | 9 skipped (3298)
```

## Checklist

- [x] Tests pass (`npm test`)


0xShadowX:
Both addressed in the follow-up commit.

- Error code: the non-JSON branch now maps the status through `statusToErrorCode(response.status, {})`, so a plain-text 429 is `RATE_LIMITED` (and 5xx keep `SERVER_ERROR`). Test `gives a plain-text 429 the same error code as a JSON one`.
- No-content-type case: added `falls back to body: null when no content type is declared and json() has consumed the body` on a real `Response` (byte body, since a string body gets a default `text/plain`). It confirms the `?? null` fallback holds and nothing throws; the body is unavailable there because `json()` has consumed it, which is the previous behaviour for that path. The older `{ json, text }` mock tests are left as they are since they cover the same fallback path.

kome12 (approved):

